PR #17, opened 2015-10-06
Scripts cleanup

Keruspe:
A trivial fix to package.json and some scripts enhancement/cleanups


Keruspe:
Note that this fixes #5 btw

PR #20, opened 2015-10-29
Display the message "build succeeded in" as a success one

BlackYoup:
The ``build succeeded in`` message appears for dedicated builds. This PR shows it as a success message.


divarvel:
I'm not okay with displaying "build succeeded" the same way as "deploy successful", since it can be confusing (people would think the deployment is over when it's not.

I'm okay with highlighting it (bold is ok, different color)

BlackYoup:
Updated, bold blue color is used

PR #21, opened 2015-10-29
Adding scale command

drouarb:


src/models/application.js
```diff
@@ -159,3 +171,31 @@ Application.redeploy = function(api, appId, orgaId) {
...
+  Logger.info("Scaling the app: " + appId);
+
+  var s_app = Application.get(api, appId, orgaId).toProperty();
+  var s_body = s_app.map(function(app) {
+    var instance = _.clone(app.instance);
+
+    instance.minFlavor = instance.minFlavor.name;
+    instance.maxFlavor = instance.maxFlavor.name;
+
+    if (scalabilityParameters.minFlavor)
```
divarvel: if the new minimum (flavor|number) is greater than the current max, then lift the maximum value to match the new minimum value.

src/commands/scale.js
```diff
@@ -0,0 +1,35 @@
...
+      params.options["min-instances"] == null && params.options["max-instances"] == null) {
+    return Logger.error("You should provide at least 1 option")
+  }
+
+  var alias = params.options.alias;
+
+  var s_appData = AppConfig.getAppData(alias);
+
+  var s_scaledApp = s_appData.flatMapLatest(function(appData) {
+    var scalabilityParameters = {};
```
divarvel: check that the provided maxes are greater than the provided mins

divarvel:
One last request, implement --instances and --flavor (can't be used at the same time as --min-* and --max-*), when you don't want auto scalability

bin/clever.js
```diff
@@ -87,6 +88,30 @@ function run() {
   var addonNameArgument = cliparse.argument("addon-name", { description: "Addon name" });
   var addonProviderArgument = cliparse.argument("addon-provider", { description: "Addon provider" });
 
+  //PARSERS
+  var flavorParser = function(flavor) {
+    var flavors = Application("listAvailableFlavors")();
+
+    if(flavors["words"].indexOf(flavor) == -1) {
```
divarvel: instead of using the autocomplete data structure, can you make Application.listAvailableFlavors return a simple list, and wrap it in words in the autocomplete handler?

divarvel:
Also, could you write tests for scalability parameters merging (between those specified as arguments and those already active for the application)

PR #29, opened 2015-11-16
Handle image variant in instance types list.

judu:
The API response to GET /products/instances will change a bit and include a new field:
```json
"variant": {
    "id": "<uuid>",
    "slug": "variant-slug",
    "name": "Variant Name"
}
```
We need to change how to check the instance availability when creating an application: we need to not only check the instance type but also the variant slug.


Keruspe:
:+1:

divarvel:
Can you please rebase master now that the travis build works?

PR #48, opened 2016-02-03
Logs spec: remove wrong parameter

BlackYoup:
Method signature is: getAppLogs(object, string, bool);


divarvel:
Oh, I've unknowingly fixed it as well.

PR #52, opened 2016-05-31
Add new node versions to test against

judu:
Just make sure that we handle new node versions.


divarvel:
Can you rebase master plz?

PR #56, opened 2016-06-08
Wait for the remote to be created before getting it

BlackYoup:


BlackYoup:
My bad, this behaviour only changed in recent version of nodegit.

PR #58, opened 2016-06-09
Rename `list` command to `applications`

### LINKED ISSUE #31: Add something like "clever aliases"
Same as "clever list" but way more intuitive as we use the word "alias" everywhere

divarvel:
`list` was not descriptive enough
Fixes #31


BlackYoup:
No deprecation message / alias to `list` for some times to keep the old behaviour ?

PR #60, opened 2016-06-10
bump dependencies

BlackYoup:
The major bump here is `lodash@^4.0.0`


BlackYoup:
Seems like tests won't pass because of a nodegit package regression : https://github.com/nodegit/nodegit/issues/1052

PR #63, opened 2016-06-10
Fix OpenBrowser and `clever open` behaviour

divarvel:
Based on #62


divarvel:
Tests were successful, but travis did not pick it up, so I had to cancel them.

PR #99, opened 2016-10-06
Webhooks management

### LINKED ISSUE #84: Support for webhooks
Webhooks can be registered on the notification API, each webhook has the following shape:

```
{
   owner_id: <user_id> or <org_id>, // the owner of the hook (user or organisation)
   urls: [ {body_format: "raw" or "slack", url: <url>], // one or more URLs to call when events happen
   scope: [<user_id or org_id or app_id or addon_id>],// the entities we're interested in
   events: [<event type>] // the events we're interested in
}
```
It's a bit hard to map that to a CLI interface without ending up with something bloated, so I think the best way is to find a subset that can be cleanly mapped to a cli interface, eg

```
# list the active notifications
clever notifications [--org <org_id>]

# create a hook with only one url
clever notifications [--org <org_id>] add \
   [--format <format>] \ # default: raw
   [--event <event_type>] \ # array, nothing means all events
   [--entity <app / addon>] \ #array, nothing means the whole owner
   <url>

# remove a hook
clever notifications [--org <org_id>] remove <hook_id>
```

It's in no way complete, but it's a good starting point. Thoughts @urcadox @BlackYoup ?

divarvel:
fixes #84

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -227,12 +227,12 @@ function run() {
   });
   notificationEventTypeOption = cliparse.option("event", {
     metavar: "type",
-    description: "Restrict notification to specific event types",
+    description: "Restrict notifications to specific event types",
     complete: Notification("listMetaEvents")
   });
   notificationScopeOption = cliparse.option("service", {
     metavar: "service_id",
-    description: "Restrict notification to specific applications and addons"
+    description: "Restrict notifications to specific applications and addons"
   });
 
   // CREATE COMMAND
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -98,9 +98,9 @@ function run() {
   var addonIdOrNameArgument = cliparse.argument("addon-id", { description: "Addon ID (or name, if unambiguous)", parser: Parsers.addonIdOrName });
   var addonNameArgument = cliparse.argument("addon-name", { description: "Addon name" });
   var addonProviderArgument = cliparse.argument("addon-provider", { description: "Addon provider" });
-  var notificationNameArgument = cliparse.argument("name", { description: "Name of the notification" });
-  var notificationHookUrlArgument = cliparse.argument("url", { description: "URL to trigger when a notification is sent" });
-  var notificationIdArgument = cliparse.argument("notification-id", { description: "Notification ID" });
+  var webhookNameArgument = cliparse.argument("name", { description: "Webhook name" });
+  var webhookUrlArgument = cliparse.argument("url", { description: "Webhook URL" });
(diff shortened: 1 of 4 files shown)


bin/clever.js
```diff
@@ -212,6 +217,23 @@ function run() {
     parser: Parsers.date,
     description: "Fetch logs after this date (ISO8601)"
   });
+  var listAllWebhooksOption = cliparse.flag("list-all", {
+    description: "List all webhooks for your user or for an organisation with the `--org` option"
+  });
+  var webhookFormatOption = cliparse.option("format", {
+    metavar: "format",
+    default: "raw",
+    description: "Format of the body sent to the webhook ('raw', 'slack', or 'flowdock')"
```
BlackYoup: "sent by the webhook" ?

bin/clever.js
```diff
@@ -212,6 +217,23 @@ function run() {
...
+  });
+  notificationEventTypeOption = cliparse.option("event", {
+    metavar: "type",
+    description: "Restrict notification to specific event types",
```
BlackYoup: Notification or Notifications ?

bin/clever.js
```diff
@@ -212,6 +217,23 @@ function run() {
...
+    complete: Notification("listMetaEvents")
+  });
+  notificationScopeOption = cliparse.option("service", {
+    metavar: "service_id",
+    description: "Restrict notification to specific applications and addons"
```
BlackYoup: same as above

src/commands/notifications.js
```diff
@@ -0,0 +1,143 @@
...
+var removeNotification = function(api, params, type) {
+  var notificationId = params.args[0];
+
+  var s_ownerId = getOrgaIdOrUserId(api, params.options.org);
+  var s_results = s_ownerId.flatMapLatest(function(ownerId) {
+    return Notification.remove(api, type, ownerId, notificationId);
+  });
+
+  s_results.onValue(function() {
+    Logger.println("The notification has been sucessfully removed");
```
BlackYoup: "successfully"

bin/clever.js
urcadox: nope, definitely sent to something but I would say "to the url" or simply "format of the body"

bin/clever.js
urcadox: +1 for notifications

bin/clever.js
urcadox: I always understood a webhook as what defines the behavior. The URL which the service posts to is the callback.

PR #115, opened 2017-02-07
drop support for node 0.12

Keruspe:
deprecated since january 2017
nodegit doesn't provide binaries for node 0.12 anymore
more and more deps are complaining that they expect >=4
this speeds up travis checks a lot


Keruspe:
Latest ubuntu LTS has node 4, debian sid too, debian wheezy doesn't even have 0.12 anyways, so it's safe to merge

PR #126, opened 2017-05-01
Add MacOS case

Korbik:
The autocomplete scripts will be added to /usr/local/share instead of /usr/share


divarvel:
poke @Keruspe

divarvel:
Thanks!

PR #136, opened 2017-06-11
Display `realId` of the add-on

k33g:
Display `realId` of the add-on when calling `clever service`: because when I call `clever env`, I can't get an environment variable for the linked fs-bucket of an application.
It's useful if you want to generate a `bucket.json` file.

Display `realId` of the add-on when calling `clever addon create ...` (same reason)

### DIFF
--- a/src/commands/addon.js
+++ b/src/commands/addon.js
@@ -68,9 +68,9 @@ var create = addon.create = function(api, params) {
 
   s_result.onValue(function(r) {
     if(linkTo) {
-      Logger.println("Addon " + name + " successfully created and linked to the application");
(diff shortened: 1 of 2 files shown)


k33g:
👋 @divarvel just a little workaround because we have not information about the bucket with the linked environment variables (yes I know, my 🇬🇧 is creepy)

k33g:
👋 @divarvel 👆

divarvel:
Yeah I'll include it in the next release

Le 17 juin 2017 7:30 AM, "Philippe Charrière" <notifications@github.com> a
écrit :

PR #140, opened 2017-08-08
drain manager

KannarFr:


src/models/drain.js
```diff
@@ -0,0 +1,136 @@
...
+    return drainTargetCredentials.username == null && drainTargetCredentials.password == null;
+}
+
+Drain.listDrainTypes = function() {
+    var typesIds = [];
+    for (var i = 0; i < Drain.types.length; i++) {
+        typesIds.push(Drain.types[i].id)
+    }
+    return autocomplete.words(typesIds);
+};
```
divarvel: It can be rewritten with map

src/models/drain.js
```diff
@@ -0,0 +1,136 @@
...
+                return Drain.credentialsEmpty(drainTargetCredentials);
+                break;
+            default:
+                return false;
+        }
+    }
+};
+
+Drain.credentialsStatus = function(drainTargetType) {
+    return Drain.types.filter(type => type.id === drainTargetType).map(type => type.credentials)[0];
```
divarvel: It would be clearer with find

src/models/drain.js
```diff
@@ -0,0 +1,136 @@
...
+    var s_res = makeJsonRequest(api, 'GET', '/logs/' + appId + '/drains', {});
+    return s_res.map(function(drains) {
+        return drains;
+    });
+};
+
+Drain.create = function(api, appId, drainTargetURL, drainTargetType, drainTargetCredentials) {
+    Logger.debug("Registering drain for " + appId);
+    if (Drain.authorizeDrainCreation(drainTargetType, drainTargetCredentials)) {
+        var body = "";
```
divarvel: you can set url and drainType first, then add (or not) credentials. Then you would be able to use const instead of var (in any case, please use let instead of var)

src/models/drain.js
```diff
@@ -0,0 +1,136 @@
...
+                return new Bacon.Error(jsonBody);
+            }
+        }
+    });
+}
+
+Drain.list = function(api, appId) {
+    Logger.debug("Fetching drains for " + appId);
+    var s_res = makeJsonRequest(api, 'GET', '/logs/' + appId + '/drains', {});
+    return s_res.map(function(drains) {
```
divarvel: is this map necessary?

KannarFr:
I applied your requested changes, I rebased and it seems I haven't to, it removed your review... Next time should I commit like "resolve divarvel review" or something like that ?

PR #141, opened 2017-08-08
It allows to display addonId and realId

k33g:
eg:
```
Addon fsbucket-acme (id:addon_b423d27f-f99f-46fa-b4e9-8d7d1abe27e7|realId:bucket_c28cf254-0134-407e-a8f1-f51505c08e61) successfully created
```

### DIFF
--- a/src/commands/addon.js
+++ b/src/commands/addon.js
@@ -68,9 +68,9 @@ var create = addon.create = function(api, params) {
 
   s_result.onValue(function(r) {
     if(linkTo) {
-      Logger.println("Addon " + name + " (id:" + r.id + "|realId:" + r.realId +") successfully created and linked to the application");
+      Logger.println("Addon " + name + " (id:" + r.id +") successfully created and linked to the application");
     } else {
-      Logger.println("Addon " + name + " (id:" + r.id + "|realId:" + r.realId +") successfully created");
+      Logger.println("Addon " + name + " (id:" + r.id +") successfully created");
     }
   });
   s_result.onError(Logger.error);
--- a/src/commands/addon.js
+++ b/src/commands/addon.js
@@ -68,9 +68,9 @@ var create = addon.create = function(api, params) {
 
   s_result.onValue(function(r) {
     if(linkTo) {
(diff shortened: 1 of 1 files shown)


divarvel:
Do you need both? It clutters the output

k33g:
I need `realId` when using fsbucket
I take the realId, replace `_` by `-` and generate this json file

```
 [{
   "bucket_host": "${id}-fsbucket.services.clever-cloud.com",
   "folder": "/storage"
 }]
```

so perhaps there are other solutions:

- output on several lines:
  ```
  Addon fsbucket-acme successfully created
  - id:addon_b423d27f-f99f-46fa-b4e9-8d7d1abe27e7
  - realId:bucket_c28cf254-0134-407e-a8f1-f51505c08e61
  ```
- a command to get the `realId` from the `id`
- or a command to get all informations of addon(s)
- or something I didn't see

and I can't make a choice between `realId` and `addon.id` because I need `addon.id when deleting

divarvel:
I see. constructing bucket host this way is fragile as the actual domain is an implementation detail.
I can add bucket_host to the environment variables exposed by the fs bucket so you don't have to craft it. That way you won't need `realId`

k33g:
@divarvel yes it's a very good solution for me
I will update the PR to only display addonId

k33g:
@divarvel if you prefer, I can delete the PR and create a new one

divarvel:
you can update the PR

divarvel:
The change has been deployed on the FS Bucket addon, new buckets provide a BUCKET_HOST env variable containing the host.

src/commands/addon.js
```diff
@@ -68,9 +68,9 @@ var create = addon.create = function(api, params) {
 
   s_result.onValue(function(r) {
     if(linkTo) {
-      Logger.println("Addon " + name + " (" + r.realId +") successfully created and linked to the application");
+      Logger.println("Addon " + name + " (id:" + r.id +") successfully created and linked to the application");
```
divarvel: Teeny tiny nitpick: can you add a space after the colon?

k33g: ok

Issue #166, opened 2017-11-30
Avoid invalid git remote names


divarvel:
Remote names are generated based on app / aliases names. In some cases that leads to invalid remote names (eg `[prod] api`).

hsablonniere:
It seems to come from and invalid range in the regex used to "slugify" the alias.

As we can see here:

https://github.com/CleverCloud/clever-tools/blob/d57871cf72f4fe50f1456d89a839ae53007a81ca/src/models/app_configuration.js#L45

The negative interval contains `A-z` which will match from uppercase `A` (ASCII-65) to  lower case `z` (ASCII-122) and this includes characters such as `[` or `]`.

divarvel:
Indeed the regex is buggy.
I thought about it, and fixing the regex is not enough, since aliases can be provided manually.

A good solution would be:
 - extract the slugifying part into its own function
 - test it
 - call the slugify function in `Git.createRemote` (or slugify its argument)

hsablonniere:
I agree about applying it directly in `Git.createRemote` but I would propose to use `_.kebabCase()` instead of doing it ourselves.

We already have `lodash` in our deps and this `kebabCase()` function has a good slugify behaviour. It would also prevent starting or ending with a dash which is not the case right now. I did a few tests and it translits characters like `é` to `e` and `ç` to `c` but it doesn't do full transliteration like `unidecode` with other characters like japanese, cyrillic...

I think we can combine both by applying `unidecode` then `_.kebabCase()`.

> extract the slugifying part into its own function

* Do we have a folder where we put this kind of utility functions?

divarvel:
+1 on `_.kebabCase()`. I think you can put it in `models/application.js`

PR #168, opened 2018-01-02
Fix slugification of app name/alias used for the git remote name

hsablonniere:
See #166 for the details.

### DIFF
--- a/src/models/app_configuration.js
+++ b/src/models/app_configuration.js
@@ -3,10 +3,10 @@ var path = require("path");
 
 var _ = require("lodash");
 var Bacon = require("baconjs");
-var unidecode = require("unidecode");
 
 var Config = require("./configuration.js");
 var Logger = require("../logger.js");
+var slugify = require("./application").slugify
 
 var AppConfiguration = module.exports = {};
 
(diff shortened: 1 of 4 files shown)


src/models/app_configuration.js
```diff
@@ -42,7 +41,7 @@ AppConfiguration.addLinkedApplication = function(appData, alias) {
     app_id: appData.id,
     deploy_url: appData.deployment.httpUrl || appData.deployment.url,
     name: appData.name,
-    alias: alias || unidecode(appData.name).replace(/[^a-zA-z0-9]+/gi, "-").toLowerCase()
+    alias: alias || appData.name
```
divarvel: can you slugify the app name in that case as well?

hsablonniere: I don't get what you mean.

The slugification of the alias happens directly in `Git.createRemote()`. When do you want to slufigy the name and what would it imply ?

divarvel: The alias is also used with the `--alias` flag. If we automatically generate an alias for the user (if they don't provide it), we make it console-friendly

divarvel: If the user specifies an alias when creating / linking the app, then we use it (we assume the user knows what they're doing)

hsablonniere: OK, so if no alias is provided through CLI, we should slugify the name and store the slugified version as an alias.

I think this means we should move the `slugify()` out of `Git.createRemote()` and back in `app_configuration.js`, WDYT?

divarvel: we still need to have it called in `Git.createRemote`, as users may pass invalid git remote names (anyway `slugify` is idempotent so it's not an issue)

divarvel: if the user wants to have a specific alias it's not an issue for CLI arguments, but for git remotes it's an issue

divarvel:
👌 thanks!

PR #169, opened 2018-01-04
Display app dependencies environment variables in `clever env`

hsablonniere:
I just tried to mimick what was being done with addon env vars. I need to dig to understand what I did.

Let me know if the wording about app dependencies is correct.

### DIFF
--- a/src/commands/env.js
+++ b/src/commands/env.js
@@ -60,12 +60,19 @@ var list = env.list = function(api, params) {
     return Env.listFromAddons(api, appData.app_id, appData.org_id);
   });
 
+  var s_env_from_deps = s_appData.flatMap(function(appData) {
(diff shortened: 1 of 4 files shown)


src/commands/env.js
```diff
@@ -60,12 +60,20 @@ var list = env.list = function(api, params) {
...
+  });
+
   var s_fullEnv = s_env.flatMapLatest(function(env) {
     return s_env_from_addons.flatMapLatest(function(env_from_addons) {
-      return {
-        manual: env,
-        addons: env_from_addons
-      };
+      return s_env_from_deps.flatMapLatest(function(env_from_deps) {
+        console.log(env)
```
divarvel: can you remove the console.log please?

hsablonniere: oups

PR #173, opened 2018-01-05
Fix "clever restart" with short commit ids

### LINKED ISSUE #167: Restarting an application with a commit string shorter than expected tries to deploy commit #Infinity
```
clever restart --commit 94e54206
Restarting <Project> on commit #Infinity
```

Using the full commit ID works as intended. I don't know if we can get the full commit ID from the short commit ID but we don't want to try to deploy the `#Infinity` commit anyway :)

hsablonniere:
* Fix wrong coercion with some short commit ids (especially those looking like an exponential)
* Resolve full commit ids through local git repo

Fixes #167

### DIFF
--- a/spec/git.spec.js
+++ b/spec/git.spec.js
@@ -97,6 +97,15 @@ describe("git", function() {
     });
   });
 
+  it("should be able to lookup a full commit id", function(done) {
+    git.resolveFullCommitId('e9e2').subscribe(function(event) {
+      expect(event.hasValue()).to.equal(true);
+      expect(event.value()).to.equal('e9e2437e73d8f10985ef8c5fd869ba9988019d80');
+      done();
+      return Bacon.noMore;
+    });
+  });
+
   it.skip("should push to the origin remote", function(done) {
     var s_push = git.getRemote("origin").flatMapLatest(function(remote) {
       return git.push(remote, "master", Bacon.once("0000000000000000000000000000000000000000"), false);
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -55,26 +55,31 @@ deploy.deploy = function(api, params) {
 };
 
 deploy.restart = function(api, params) {
-  var alias = params.options.alias;
-  var quiet = params.options.quiet;
-  var commitId = params.options.commit;
-  var withoutCache = params.options["without-cache"];
+  const { alias, quiet, commit, 'without-cache': withoutCache } = params.options
 
-  var s_appData = AppConfig.getAppData(alias).toProperty();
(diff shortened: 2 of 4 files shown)


hsablonniere:
I'm not so sure about the unit test for `Git.resolveFullCommitId()` since it relies on an external test repo. WDTY?

hsablonniere:
I also applied the boyscout mantra so I reworked the `restart` function a bit with some ES6!

src/commands/deploy.js
```diff
@@ -55,26 +55,31 @@ deploy.deploy = function(api, params) {
...
-    Logger.println("Restarting " + app_data.name + suffix);
-    return Application.redeploy(api, app_data.app_id, app_data.org_id, commitId, withoutCache);
-  }).toProperty();
+  const { alias, quiet, commit, 'without-cache': withoutCache } = params.options
+
+  const s_fullCommitId = Git.resolveFullCommitId(commit)
+  s_fullCommitId.onError((e) => {
+      Logger.error(`Commit id ${commit} is ambiguous`)
+      process.exit(1);
+    })
```
divarvel: I try to avoid early returns, would it be possible to return an error stream and let the error be managed only at the end?

src/commands/deploy.js
```diff
@@ -55,26 +55,31 @@ deploy.deploy = function(api, params) {
...
-    var suffix = "";
-    if(commitId) suffix += " on commit #" + commitId;
-    if(withoutCache) suffix += " without using cache";
-
-    Logger.println("Restarting " + app_data.name + suffix);
-    return Application.redeploy(api, app_data.app_id, app_data.org_id, commitId, withoutCache);
-  }).toProperty();
+  const { alias, quiet, commit, 'without-cache': withoutCache } = params.options
+
+  const s_fullCommitId = Git.resolveFullCommitId(commit)
```
divarvel: what happens if no commit is given?

hsablonniere: It fails :scream: (fixing in progress...)

spec/git.spec.js
```diff
@@ -97,6 +97,16 @@ describe("git", function() {
     });
   });
 
+  it("should be able lookup a full commit id", function(done) {
```
divarvel: "to lookup"

src/models/git.js
```diff
@@ -184,6 +184,18 @@ module.exports = function(repositoryPath) {
     });
   };
 
+  Git.resolveFullCommitId = function(commitId) {
+    if (commitId == null) {
+      return Bacon.constant(null);
+    }
```
divarvel: I'd prefer to have the null checked outside of this function

PR #174, opened 2018-01-05
Add "clever profile" command

### LINKED ISSUE #161: Add a command to show the current logged user profile
It would tell the user some information about its account, maybe: name, email, 2FA status (and a link to the console to enable it if it's not ?)

For example, npm uses `npm profile` to get the user's profile

hsablonniere:
Fixes #161

No unit test, do we need one for this kind of stuffs?

Looks like this:

capture d ecran_2018-01-05_18-28-40


divarvel:
I think we can skip the unit tests for this.

divarvel:
Thanks!

PR #175, opened 2018-01-08
Show commit about to be redeployed

### LINKED ISSUE #145: clever restart: show commit about to be redeployed
What do you think ? Since we can deploy any old pushed commit of the tree, it's not always HEAD.

Maybe, we could even show the associated commit message

hsablonniere:
This is a quick fix, it may deserve some refactoring on the way we use the `s_commitId` Bacon stream.

Fixes #145.

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -57,7 +57,7 @@ deploy.deploy = function(api, params) {
 deploy.restart = function(api, params) {
   const { alias, quiet, commit, 'without-cache': withoutCache } = params.options
 
-  const s_fullCommitId = (commitId == null) ?
+  const s_fullCommitId = (commit == null) ?
     Bacon.constant(null) :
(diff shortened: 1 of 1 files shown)


src/commands/deploy.js
```diff
@@ -68,12 +68,12 @@ deploy.restart = function(api, params) {
...
 
   var s_deploy = s_appData.flatMapLatest(function(app_data) {
-    var suffix = "";
-    if(commitId) suffix += " on commit #" + commitId;
-    if(withoutCache) suffix += " without using cache";
-
-    Logger.println("Restarting " + app_data.name + suffix);
-    return Application.redeploy(api, app_data.app_id, app_data.org_id, commitId, withoutCache);
+    return s_commitId.flatMapLatest((theCommitId) => {
+      var suffix = " on commit #" + (commitId || theCommitId);
```
divarvel: can you use let instead of var?

PR #176, opened 2018-01-08
Add key param for ssh command

### LINKED ISSUE #164: Add option on `clever ssh` for custom ssh config
From what I understood, `clever ssh` uses by default the `id_rsa` one.
Would be nice to have a way to give an option to make it use my custom clever ssh config.

That's all! Thanks for your time folks! 👋

hsablonniere:
I took the opportunity to refactor/cleanup the ssh command a bit...

Fixes #164.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -283,6 +283,11 @@ function run() {
     metavar: "addon_id",
     description: "Addon ID"
   });
+  var sshIdentityFileOption = cliparse.option("identity-file", {
+    aliases: ["i"],
+    metavar: "identity-file",
+    description: "SSH Identity file",
+  });
 
   // CREATE COMMAND
   var appCreateCommand = cliparse.command("create", {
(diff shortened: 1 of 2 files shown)


src/commands/deploy.js
```diff
@@ -68,12 +68,12 @@ deploy.restart = function(api, params) {
...
 
   var s_deploy = s_appData.flatMapLatest(function(app_data) {
-    var suffix = "";
-    if(commitId) suffix += " on commit #" + commitId;
-    if(withoutCache) suffix += " without using cache";
-
-    Logger.println("Restarting " + app_data.name + suffix);
-    return Application.redeploy(api, app_data.app_id, app_data.org_id, commitId, withoutCache);
+    return s_commitId.flatMapLatest((theCommitId) => {
+      var suffix = " on commit #" + (commitId || theCommitId);
```
divarvel: can you use `let` instead?

hsablonniere: I'll remove it when I'll rebase on master.

PR #177, opened 2018-01-09
Only use console.log & console.error

### LINKED ISSUE #134: `console.debug` eats logging output in node 8
Before node 8, `console.debug` did not exist, so the logger defaulted to `console.log`.
In node 8, `console.debug` exists, but does not log its arguments (I guess it can be done with an env variable or something).

It's not documented yet (https://t.co/JHrEGE3kT3),  so I don't know how to tell it to display logs when `CLEVER_VERBOSE` is set.

hsablonniere:
This fixes remove usage of `console.debug` and `console.warn` to use `console.log` for everything except for errors (where we use `console.error`).

I also took the liberty to refactor some sutffs:

* Add `use_strict`
* Use some ES6+ goodies like spread, const or arrow functions
* Remove bind calls since `console` is a namespace like `Math` or `JSON`

More infos on the "autobind" behaviour:

* https://bugs.chromium.org/p/chromium/issues/detail?id=167911
* https://bugs.webkit.org/show_bug.cgi?id=157286
* https://github.com/whatwg/console/issues/3#issuecomment-254363388
* https://console.spec.whatwg.org/#console-namespace
* https://wpt.fyi/console

FYI, it works on Chrome, Node, Safari, FF but not yet in Edge.

Fixes #134.

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,36 +1,33 @@
-var _ = require("lodash");
+'use strict'
 
-var processApiError = function(error) {
-  if(error.id && error.message) {
-    var line = error.message + " [" + error.id + "]";
-    if(error.fields) {
-      line += "\n" + Object.keys(error.fields).map(function(field) {
-        return field + ": " + error.fields[field];
-      }).join("\n");
+const _ = require('lodash')
+
+const processApiError = function (error) {
+  if (error.id == null || error.message == null) {
+    return error
+  }
+  const fields = _.map(error.fields, (msg, field) => `${field}: ${msg}`)
+  return [`${error.message} [${error.id}]`, ...fields].join('\n')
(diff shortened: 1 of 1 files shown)


src/logger.js
```diff
@@ -1,36 +1,33 @@
...
-    logger[severity] = function() {};
+const processApiError = function (error) {
+  if (error.id == null || error.message == null) {
+    return error
   }
-  return logger;
-}, module.exports);
+  const fields = _.map(error.fields, (msg, field) => `${field}: ${msg}`)
+  return [`${error.message} [${error.id}]`, ...fields].join('\n')
+}
```
divarvel: That's way cleaner indeed 👍

PR #200, opened 2018-03-07
Fix bad libcurl on linux build

### LINKED ISSUE #194: linux packaging: Use the nodegit we just built
The "official" nodegit uses libcurl-gnutls.so which is non standard and doesn't work everywhere.

hsablonniere:
Fixes #194

### DIFF
--- a/nodegit/nodegit.md
+++ b/nodegit/nodegit.md
@@ -1,10 +1,21 @@
-Those modules can be found here:
+# nodegit.node archives 
+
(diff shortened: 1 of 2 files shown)


divarvel:
Looks good to me, the resulting binary has been tested on arch and exherbo systems, right?

hsablonniere:
Yep but we'll check again with the other bugfixes/features just before the release.

PR #206, opened 2018-03-15
Issue 202 error handling small commands

hsablonniere:
Refactor for:

* profile
* makeDefault
* ssh
* applications


divarvel:
Oo a travis build issue that's actually a real issue!

divarvel:
So the failure is that `conf.CONFIGURATION_FILE` is not used and the conf is read from `~/.config/clever-cloud`

hsablonniere:
I forgot to run them locally :-(

hsablonniere:
Fixed and updated!

PR #216, opened 2018-04-06
add desktop notification when deploy succeeds or fails

ldoguin:


hsablonniere:
Hey @ldoguin, I was about to ask you to rework this according to the new master (which had a huge ES6+ refactoring) but I think we should consider use cases. With your current code, every user would get notifications.

1. Should we let user enable/disable those desktop notifications?
2. Should this feature be opt-in or opt-out?
3. How would this work for CI (or headless) environment?

hsablonniere:
Here's how I see it:

1. Yes we should let user enable/disable it. We can store this information at the same place we store the token and secret (~/.config/clevercloud). This mean it would be global for all projects of a given user system.
2. I would prefer opt-in so people don't get annoyed by it without consent.
3. If it's opt-in, we don't have to think about CI environment, it would be disabled by default.

hsablonniere:
@ldoguin We'll rediscuss this one day.

PR #231, opened 2018-06-17
Avoid useless loading of nodegit

fmenou:
Explicit require of nodegit when we don't need it forces us to have nodegit properly built on our target platform (which can be tricky) even when we don't actually use it (we don't use `clever deploy` but instead direct git push).

I'm not sure if
 - that make sense
 - that actually fixes my problem
 - the resulting archive is loadable (couldn't find a way to test it)

Please advice :pray:

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -1,12 +1,10 @@
 var _ = require("lodash");
 var path = require("path");
 var Bacon = require("baconjs");
-var nodegit = require("nodegit");
 
 var Logger = require("../logger.js");
 
 var Application = require("../models/application.js");
-var Git = require("../models/git.js")(path.resolve("."));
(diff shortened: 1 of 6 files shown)


ptitfred:
If I was given a linux64 bit tar gz for that branch I'm willing to test it on my side before one would merge it.

hsablonniere:
Hey @ptitfred, we're very close to completely remove nodegit and replace it with a full JS alternative.

I think it would solve your problem, WDYT?

More details here: https://github.com/CleverCloud/clever-tools/issues/236

ptitfred:
I've solved the core problem by moving my setup to nix. This PR is therefore pure code cleanup. Up to you to accept or close it :)

Issue #247, opened 2018-08-09
Open the Clever Cloud Console with a command


cnivolle:
This command would be like `clever open` but for the web console, not the URL of the application itself.
I was wondering if this kind of command could be useful to anyone, other than me.

The usage would be quite simple: 
Open the web console directly to the app's informations, with a command like `clever console` (maybe with using the option `--alias` to route to the correct app URL).

I'm open to anything smarter than `clever console`.

divarvel:
+1 that sounds reasonable.

`clever console [--alias <alias>]` is great

Le jeu. 9 août 2018 15:41, Clément Nivolle <notifications@github.com> a
écrit :

hsablonniere:
Fixed by @grallc!! Thanks again :tada:
As I mentioned this feature will be shipped with the new release.
It may take a bit of time still.

hsablonniere:
@grallc Your feature is present in the beta versions ;-)

PR #248, opened 2018-08-20
Added console command #247

grallc:
Added console command (#247) to open the application's console from CLI and removed dupllicated addonName arg

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -91,7 +91,6 @@ function run () {
       parser: Parsers.addonIdOrName,
     }),
     addonName: cliparse.argument('addon-name', { description: 'Addon name' }),
-    addonName: cliparse.argument('addon-name', { description: 'Addon name' }),
     addonProvider: cliparse.argument('addon-provider', { description: 'Addon provider' }),
     alias: cliparse.argument('app-alias', { description: 'Application alias' }),
     appIdOrName: cliparse.argument('app-id', {
@@ -501,6 +500,13 @@ function run () {
     options: [opts.alias],
   }, open);
 
+    // CONSOLE COMMAND
+    const consoleModule = lazyRequireFunctionWithApi('../src/commands/console.js');
+    const consoleCommand = cliparse.command('console', {
+      description: 'Open an application in the console',
+      options: [opts.alias],
+    }, consoleModule);
+
   // PROFILE COMMAND
   const profile = lazyRequireFunctionWithApi('../src/commands/profile.js');
   const profileCommand = cliparse.command('profile', {
@@ -631,6 +637,7 @@ function run () {
       logsCommand,
       makeDefaultCommand,
       openCommand,
+      consoleCommand,
       profileCommand,
       publishedConfigCommands,
       restartCommand,
--- a/src/commands/console.js
+++ b/src/commands/console.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const AppConfig = require('../models/app_configuration.js');
+const handleCommandStream = require('../command-stream-handler.js');
+const Logger = require('../logger.js');
+const OpenBrowser = require('../open-browser.js');
+
+function consoleModule (api, params) {
+  const { alias } = params.options;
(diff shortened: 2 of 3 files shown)


hsablonniere:
@grallc Hey Corentin, thanks very much for you contribution :tada:

* I have a few tiny remarks but it looks very good :wink:
* FYI: The release of this code may take some time since we're in the middle of a big change on the releasing process...

bin/clever.js
```diff
@@ -501,6 +500,13 @@ function run () {
     options: [opts.alias],
   }, open);
 
+    // CONSOLE COMMAND
+    const console = lazyRequireFunctionWithApi('../src/commands/console.js');
```
hsablonniere: I'm afraid this naming could confuse people with the already existing global variable `console`.

* Could you rename this?
* Maybe something like `consoleModule` or `consoleFn`

src/commands/console.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const AppConfig = require('../models/app_configuration.js');
+const Domain = require('../models/domain.js');
+const handleCommandStream = require('../command-stream-handler.js');
+const Logger = require('../logger.js');
+const OpenBrowser = require('../open-browser.js');
+
+function console (api, params) {
```
hsablonniere: Same here :smile:

I'm afraid this naming could confuse people with the already existing global variable `console`.

* Could you rename this?
* Maybe something like `consoleModule` or `consoleFn`

src/commands/console.js
```diff
@@ -0,0 +1,20 @@
...
+  const { alias } = params.options;
+  const s_open = AppConfig.getAppData(alias)
+    .flatMapLatest(({ app_id, org_id }) => {
+      Logger.println('Opening the console in your browser');
+      return OpenBrowser.openPage('https://console.clever-cloud.com/users/me/applications/' + app_id);
```
hsablonniere: I haven't tested but I think it won't work for apps in organizations.
We miss a test on the `org_id` to build the right URL in any case.

I also think we should use template strings for this:

```js
const appUrl = `https://console.clever-cloud.com/users/me/applications/${app_id}`;
```

grallc:
Just pushed the amended commit.
Sorry for the organisation, I didn't know their existence... It's fixed now!
Thanks you for your tips ! :smile:

src/commands/console.js
```diff
@@ -0,0 +1,24 @@
...
+  const { alias } = params.options;
+  const s_open = AppConfig.getAppData(alias)
+    .flatMapLatest(({ app_id, org_id }) => {
+      Logger.println('Opening the console in your browser');
+      let path;
+
+      if (org_id) {
+        path = `organisations/${org_id}`;
+      }
+      return OpenBrowser.openPage(`https://console.clever-cloud.com/${path || 'users/me'}/applications/${app_id}`);
```
hsablonniere: What do you think of this?

```js
const path = (org_id != null) ? `organisations/${org_id}` : 'users/me';
return OpenBrowser.openPage(`https://console.clever-cloud.com/${path}/applications/${app_id}`);
```

grallc: Oh thanks!
I didn't know this syntax :sweat_smile:

package.json
```diff
@@ -13,7 +13,7 @@
   "author": "Rodolphe Belouin <rodolphe.belouin@gmail.com>",
   "license": "MIT",
   "bin": {
-    "clever": "./bin/clever.js",
+    "clever-test": "./bin/clever.js",
```
hsablonniere: Ooops

grallc: One more...
Really sorry for the spam :(

hsablonniere: This is not spam. This is iterative contribution :wink: which is always welcome.

hsablonniere:
@grallc I'll test this in the afternoon and merge it if everything is OK.

hsablonniere:
@grallc I just merged your commit manually (so I could remove the diff on the `package-lock.json`. Thanks again for you contribution!!!

:tada:

PR #261, opened 2018-10-11
add "clever version"

Keruspe:
Currently, we have `--version` but we do not have a `version` subcommand which is quite common in projets using subcommands like us


hsablonniere:
Yep, good idea. As specified by section 3 of this article:

https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46

PR #298, opened 2018-12-22
fix(login): fix login issue on macos

DecampsRenan:
**TLDR; : https://github.com/sindresorhus/opn/issues/91**

The `login` command never end successfull (infinite wait)  on mac due to an issue with the `open` command used by the `opn` package; the promise returned when the `opn` function is executed was returned only when the browser was closed.

Fixed by passing the `{ wait: false }` option param.

I haven't tested the fix on linux / windows; so I don't know if it break something else 😉


hsablonniere:
Hey @DecampsRenan ;-)

Thanks very much for this!!!

I took your commit and:
* removed the modifications of the `package-lock.json`
* rebased it on latest master

Issue #296, opened 2018-11-30
WebSocket closes automatically after 500ms


haitlahcen:
As stated [here](https://github.com/CleverCloud/clever-tools/blob/master/src/models/ws-stream.js#L70), when a `WebSocket` stream is opened, it is infinite. 

This is very problematic in case of network issues (i.e. the CLI will never exit). 

Maybe we should consider having a limit for the number of retry with an incremental timeout for the re-connection.

clementd-fretlink:
While investigating on #294, we've been able to get a bit more information (by adding some logs in `ws-stream.js`). It took some time because we had to deploy a custom version with extra logs, and our CI environment makes it a bit difficult to get logs, but we got them eventually.

So, the current state of affairs is:
- the websocket closes "normally" (we get a `Normal Closure` event in the `close` event handler)
- judging by the timestamp, this event is received every 2s seconds. Since the reconnection timeout is 1.5s, it seems to indicate that the connection closes after 500ms.

This would expain the "not stopping" behaviour, since the websocket is closed ~75% of the time, it misses events.

With that in mind, I have two questions:
- do you have more information in the server logs?
- would getting events through SSE instead of websockets be feasible?

hsablonniere:
@clementd-fretlink @haitlahcen Thanks for the details, it helps a lot.

clementd-fretlink:
(I can submit a PR with the more detailed logs if you want)

hsablonniere:
(I'm currently adding logs locally to get a better understanding of what's going on but please do, it can only help me a bit more)

divarvel:
We're running one last test, to make sure the WS is not closed by the cleanup function of the observable (it's the only place `ws.close` is called, so…)

divarvel:
Ok, so we are now sure that it isn't the client that triggers the close.
@hsablonniere could you describe what happens when the auth fails? Is the server closing the connection directly? Does it send a message in the socket describing the error?

hsablonniere:
I tried with a random string for the auth to force a bad auth. I got this:

```
ENTER openStream(/*...*/)
ENTER openWebSocket(/*...*/)
ENTER new WebSocket(/*...*/)
[DEBUG] Websocket opened successfully: wss://api.clever-cloud.com/v2/events/event-socket
[DEBUG] Websocket closed.
[WARN] Websocket closed, reconnecting
CLEANUP CLOSE
```

hsablonniere:
[...]

PR #300, opened 2019-01-10
Limit the number of retries for when the WebSocket connexion fails

### LINKED ISSUE #294: `clever deploy` not exiting on deployment success
We are migrating our CI to `v1.1.1` and we observed that `clever deploy` did not exit upon deployment success.

Looking at the code in the stream command handler, I see an explicit `process.exit` for errors, but not for the end of the stream. [0]

I've witnessed a few times with bacon that sometimes the process keeps running even though we've got to the end of a stream. I guess it's because there is still an active event listener registered somewhere.

In the case of `clever deploy` (or `clever restart`), i feel that an explicit exit at the end of the deployment would be safer_

[0]: https://github.com/CleverCloud/clever-tools/blob/master/src/command-stream-handler.js#L6

hsablonniere:
Fixes #294 & #296

### DIFF
--- a/src/models/ws-stream.js
+++ b/src/models/ws-stream.js
@@ -5,20 +5,23 @@ const WebSocket = require('ws');
 
 const Logger = require('../logger.js');
 
+const RETRY_DELAY = 1500;
+const MAX_RETRY_COUNT = 5;
+const PING_INTERVAL = 40000;
+
 function openWebSocket (url, authorization) {
   return Bacon.fromBinder((sink) => {
     const ws = new WebSocket(url);
     let pingInterval;
 
     ws.on('open', () => {
-      Logger.debug('Websocket opened successfully: ' + url);
+      Logger.debug('WebSocket opened successfully: ' + url);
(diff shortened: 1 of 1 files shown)


src/models/ws-stream.js
```diff
@@ -31,13 +34,13 @@ function openWebSocket (url, authorization) {
...
     ws.on('close', () => {
-      Logger.debug('Websocket closed.');
+      Logger.debug('WebSocket closed.');
       clearInterval(pingInterval);
       sink(new Bacon.End());
     });
 
     ws.on('error', () => {
-      Logger.debug('Websocket closed.');
+      Logger.debug('WebSocket closed.');
```
clementd-fretlink: Maybe log the error and put a different message to distinguish between normal closing and errors? Having the same messages made debugging more difficult

hsablonniere: Good catch.

hsablonniere:
@clementd-fretlink I also returned a `Bacon.Error` in the sink.

PR #307, opened 2019-02-27
fix(activity): avoid using cursor when stdout is not a tty

### LINKED ISSUE #306: Clever activity fails when stdout is not a tty
The function `moveCursor` from `process.stdout.moveCursor` is undefined when `stdout` is not a tty:

`TypeError: process.stdout.moveCursor is not a function\n    at clearPreviousLine (/snapshot/clever-tools_1.3.0/src/commands/activity.js:62:18)`

This behavior can be reproduced with:
`node -p -e "process.stdout.moveCursor(0, -1)" `
`node -p -e "process.stdout.moveCursor(0, -1)" | cat`

Fortunately, node is exposing a `isTTY` [attribute](https://nodejs.org/api/tty.html#tty_tty)

haitlahcen:
This PR will hopefully fix #306

### DIFF
--- a/src/commands/activity.js
+++ b/src/commands/activity.js
@@ -59,9 +59,11 @@ function isTemporaryEvent (ev) {
 }
 
 function clearPreviousLine () {
-  process.stdout.moveCursor(0, -1);
-  process.stdout.cursorTo(0);
-  process.stdout.clearLine();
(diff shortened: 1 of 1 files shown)


hsablonniere:
Thanks very much @haitlahcen

PR #308, opened 2019-03-08
remove datadog tcp creation drain, add new datadog http drain creation

KannarFr:
We keep the management of them, but creation is now unavailable.


KannarFr:
@hsablonniere ready to merge and release! :)

PR #311, opened 2019-03-21
Remove leftover `console.log` in `clever env import`

clementd-fretlink:


hsablonniere:
Thanks @clementd-fretlink for the fix :+1:

PR #317, opened 2019-04-03
add new 'clever env import-vars' command

### LINKED ISSUE #316: Add the ability to publish environment variables from the current shell.
Let's say my application needs some environment variables, namely `FOO` and `BAR`.
For testing purposes I exported those in my shell in which I'm testing my app.
I would love to just do `clever export-env FOO BAR` to automatically create those variables on Clever Cloud matching the ones in my shell.

Keruspe:
Fixes #316

### DIFF
--- a/src/commands/env.js
+++ b/src/commands/env.js
@@ -67,4 +67,25 @@ async function importEnv (params) {
   Logger.println('Environment variables have been set');
 };
 
-module.exports = { list, set, rm, importEnv };
+async function importEnvVars (params) {
+  const [envNames] = params.args;
+  const { alias } = params.options;
+
+  for (const envName of envNames) {
+    const nameIsValid = validateName(envName);
+    if (!nameIsValid) {
+      throw new Error(`Environment variable name ${envName} is invalid`);
+    }
+  }
+
+  const { org_id, app_id: appId } = await AppConfig.getAppData(alias).toPromise();
+
+  for (const envName of envNames) {
+    const value = process.env[envName] || '';
+    await application.updateEnvVar({ id: org_id, appId, envName }, { value }).then(sendToApi);
+  }
+
+  Logger.println('Your environment variables have been successfully saved');
+};
+
+module.exports = { list, set, rm, importEnv, importEnvVars };
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -143,6 +143,10 @@ function run () {
     notificationId: cliparse.argument('notification-id', { description: 'Notification ID' }),
     webhookUrl: cliparse.argument('url', { description: 'Webhook URL' }),
     envVariableName: cliparse.argument('variable-name', { description: 'Name of the environment variable' }),
+    envVariableNames: cliparse.argument('variable-names', {
+      description: 'Comma separated list of names of the environment variables',
+      parser: Parsers.commaSeparated,
+    }),
     envVariableValue: cliparse.argument('variable-value', { description: 'Value of the environment variable' }),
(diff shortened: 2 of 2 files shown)


divarvel:
That's a neat feature, but naming the command `export` is inconsistent with the current behaviour: `import` is used to update clever cloud env variables (and so `export` is implicitly associated to updating local env based on the env on clever-cloud)

divarvel:
Additionally, a comma-separated list of args is :-/ (but that's a limitation of cliparse-node, afaik, so this one's on me :hugs: )

Keruspe:
yeah, the comma-separated part is because of cliparse, there's no "all the other args" option.

I was just fixing the name as you were commenting on it :)

divarvel:
:ok_hand: that's really neat!

Keruspe:
@hsablonniere rebased

Keruspe:
Fixed the commit message

src/commands/env.js
```diff
@@ -67,4 +67,25 @@ async function importEnv (params) {
...
+async function importEnvVars (params) {
+  const [envNames] = params.args;
+  const { alias } = params.options;
+
+  for (const envName of envNames) {
+    const nameIsValid = validateName(envName);
+    if (!nameIsValid) {
+      throw new Error(`Environment variable name ${envName} is invalid`);
+    }
+  }
```
hsablonniere: Do we need to validate env var names here?

I mean, we're about to try to read them from current env so if they aren't valid, they won't be in the current env.

I guess it's a good way to fail fast and help a user that maybe made a typo in the command.

I say OK but maybe we would like to throw an error with all the bad names.

src/commands/env.js
```diff
@@ -67,4 +67,25 @@ async function importEnv (params) {
...
+
+  const { org_id, app_id: appId } = await AppConfig.getAppData(alias).toPromise();
+
+  for (const envName of envNames) {
+    const value = process.env[envName] || '';
+    await application.updateEnvVar({ id: org_id, appId, envName }, { value }).then(sendToApi);
+  }
```
hsablonniere: I'd rather have one call with all variables at once.

Right now it's not really a problem because if you have 3 vars and if the 3rd one fails, you run the command again and since `updateEnvVar` is idempotent so it's no big deal.

Having one call will be a bit faster and will have just one opportunity for network failure.

src/commands/env.js
Keruspe: The idea here was to stay as close to importEnv as possible, and if we're launched from somewhere else than a shell, the env might not be as strict as we think

src/commands/env.js
Keruspe: There's nothing in the API for that, though, ATM, and I'm not sure we want to add such an endpoint just for this, as it would be confusing VS the "override all the env" one.
We could fetch the whole env, edit it and then repush it but isn't that actually more overkill?

src/commands/env.js
hsablonniere: Good catch.

src/commands/env.js
hsablonniere: There is ;-)

We have this:

```
PUT /organisations/{id}/applications/{appId}/env
PUT /self/applications/{appId}/env
```

It's used by the CLI and the console.

hsablonniere: OK, again, I just understood my mistake.

hsablonniere:
Merged locally, thanks @Keruspe ;-) :clap:

PR #326, opened 2019-06-28
Import environment variables from a JSON value

clementd-fretlink:
# Context

In automated environments, the `env`-like format requires manual generation and is error-prone.
Reading directly a JSON value allows easier plugging in CI setups, as well as eliminating corner cases (empty vars, multiline vars, etc).

# Changes

This PR adds a `--json` flag to `clever env import`, allowing to read values in JSON format.

# Questions

For symmetry, I could add json support in `clever env export` as well

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -131,6 +131,7 @@ function run () {
       aliases: ['f'],
       description: 'Display access logs continuously (ignores before/until, after/since)',
     }),
+    importJSON: cliparse.flag('json', { description: 'Import env variables from a JSON value' }),
     addonId: cliparse.option('addon', { metavar: 'addon_id', description: 'Addon ID' }),
     after: cliparse.option('after', {
       metavar: 'after',
@@ -495,6 +496,7 @@ function run () {
   }, env('rm'));
   const envImportCommand = cliparse.command('import', {
     description: 'Load environment variables from STDIN\n(WARNING: this deletes all current variables and replace them with the new list loaded from STDIN)',
+    options: [opts.importJSON]
   }, env('importEnv'));
   const envImportVarsCommand = cliparse.command('import-vars', {
     description: 'Add or update environment variables named <variable-names> (comma separated), taking their values from the current environment',
--- a/src/commands/env.js
+++ b/src/commands/env.js
@@ -58,10 +58,11 @@ async function rm (params) {
 };
 
 async function importEnv (params) {
-  const { alias } = params.options;
+  const { alias, json } = params.options;
+  const format = json ? 'json' : 'env';
   const { ownerId, appId } = await AppConfig.getAppDetails({ alias });
 
-  const vars = await variables.readVariablesFromStdin();
+  const vars = await variables.readVariablesFromStdin(format);
   await application.updateAllEnvVars({ id: ownerId, appId }, vars).then(sendToApi);
 
   Logger.println('Environment variables have been set');
--- a/src/models/variables.js
+++ b/src/models/variables.js
@@ -1,7 +1,8 @@
 'use strict';
 
+const _ = require('lodash');
 const readline = require('readline');
-const { ERROR_TYPES, parseRaw, toNameValueObject } = require('@clevercloud/client/cjs/utils/env-vars.js');
+const { ERROR_TYPES, parseRaw, toNameValueObject, validateName } = require('@clevercloud/client/cjs/utils/env-vars.js');
 
 function readStdin () {
 
@@ -28,17 +29,28 @@ function readStdin () {
 }
 
 function parseFromJSON (rawStdin) {
+  let json;
   try {
-    const json = JSON.parse(rawStdin);
-    // The JSON structure is not checked before sending it to the API.
-    // In the case of an error, the API call will fail and an error will
-    // be logged. It could be possible to add a check here to improve the
-    // UX a bit (but we'd have to make sure it's consistent with what the
-    // API checks).
-    return json;
+    json = JSON.parse(rawStdin);
   }
   catch (e) {
     throw new Error('Error when parsing json', e);
+  }
+
+  if (_.isArray(json)) {
+    const isValidPair = ({ name }) => validateName(name);
+
+    const invalidPairs = json.filter((p) => !isValidPair(p));
+
+    if (invalidPairs.length !== 0) {
+      throw new Error(invalidPairs.map(({ name }) => `${name} is not a valid variable name`));
+    }
+    else {
+      return toNameValueObject(json);
+    }
+  }
+  else {
+    throw new Error('You need to provide a list of key / value pairs, for instance: [{"name": "PORT", "value": 8080}]');
   }
 }
 
--- a/src/models/variables.js
+++ b/src/models/variables.js
@@ -27,9 +27,22 @@ function readStdin () {
   });
 }
 
-async function readVariablesFromStdin () {
+function parseFromJSON (rawStdin) {
+  try {
+    const json = JSON.parse(rawStdin);
+    // The JSON structure is not checked before sending it to the API.
+    // In the case of an error, the API call will fail and an error will
+    // be logged. It could be possible to add a check here to improve the
+    // UX a bit (but we'd have to make sure it's consistent with what the
+    // API checks).
+    return json;
+  }
+  catch (e) {
+    throw new Error('Error when parsing json', e);
+  }
+}
 
-  const rawStdin = await readStdin();
+function parseEnvLines (rawStdin) {
   const { variables, errors } = parseRaw(rawStdin);
 
   if (errors.length !== 0) {
@@ -55,6 +68,20 @@ async function readVariablesFromStdin () {
   }
 
   return toNameValueObject(variables);
+}
+
+async function readVariablesFromStdin (format = 'env') {
+
+  const rawStdin = await readStdin();
+
+  switch (format) {
+    case 'env':
+      return parseEnvLines(rawStdin);
+    case 'json':
+      return parseFromJSON(rawStdin);
+    default:
+      throw new Error("Unrecognized environment input format. Available formats are 'env' and 'json'");
+  }
 }
 
 module.exports = {


hsablonniere:
I looked at your code.

I'm currently integrating the new parser/validator that we created for the new env vars Web Components.

The thing that bothers me right now (not related to your PR) is that `clever env` is not really a the equivalent of `clever env import` because `clever env` lists env vars from app, addons and linked apps.

hsablonniere:
I missed that but your code seems to propose this:

```json
{
	"VAR_ONE": "ONE",
	"VAR_TWO": "TWO"
}
```

I want to discuss the cases because I'd like to consider:

```json
[
	{ "name": "VAR_ONE": "value": "ONE" },
	{ "name": "VAR_TWO": "value": "TWO" }
]
```

divarvel:
Indeed, it uses an object and not an associative list, it seemed to me that
was the most natural way (as it matches more closely the env semantics and
`process.env` in node), but either way could work imo. What's bothering you
with an object?

How would you handle duplicate keys in the associative list?

hsablonniere:
* What's your use case?
* Who produces the JSON you import?
  * Written by a human?
  * Produced by a program?

I'm still considering both but for the duplicates, I would reuse the same system as the `NAME=VALUE`:

Capture d’écran_2019-07-12_15-03-18

I'm considering adding those:

* `clever env --app` (only env vars from app)
* `clever env --addons` (only env vars from addons)
* `clever env --linked-apps` (only env vars from linked apps)

This needs real thinking so I'll move the discussion in an issue.
This way we could not break `clever env` and let it produce everything, just like it does right now.

It would be easy to support `--json` for all those cases, the question is, what do we do for `clever env`?

I'll create the issue and we'll try to solve this and ship it in the next few weeks.

In the meantime, I will ship the new parser which supports multiline strings etc...

divarvel:
The use case would be generating an env file from ansible. For
machine-to-machine communication, I'd rather have json, which is well
defined, instead of a custom format with a hand-rolled serializer.

I don't really care about the actual json schema (as long as it's faithful
to the thing we model: the env variables).

As for `clever env`, I think it deserves its own discussion

clementd-fretlink:
Is there anything you want me to change in the current PR?

clementd-fretlink:
:wave: did you find a satisfying solution for the env splitting?

celine-m-s:
I'd like to upvote this PR. Is this something you still have in mind?
I'm currently struggling to update the environment variables of my app with the CLI. Passing a JSON would be just perfect.

divarvel:
Generating a JSON file is way easier than generating a custom format, especially when escaping is involved.

I'm still available to update this PR as needed :slightly_smiling_face:

clementd-fretlink:
I've rebased the PR on the latest master and made sure it works. I think the error handling for the JSON import could be improved but i'm not sure how thorough you want the check to be.

IMO checking that the input is an object with only scalar values would be a good thing

clementd-fretlink:
Hi again! This issue is starting to get quite annoying on our side and requires a lot of kludges in our CI. What can I do to help this feature move forward?

hsablonniere:
Hey, sorry for the delay, as you know, we're always struggling on priorities between differents codebases (client, components, CLI, console...) with limited resources.

I deferred this issue because I naïvely remembered it required a big refactoring for the export but let's focus on the import first...

I will have a look at your PR this afternoon.

clementd-fretlink:
Sure, thanks! This PR only concerns the import indeed. I tried to keep changes as minimal as possible

hsablonniere:
OK, it seems almost OK :wink:

I like your comments about the validation but I think we should add something after the JSON parsing to reuse some client-side validators we already have.

* For the name validation, `@clevercloud/client/cjs/utils/env-vars.js` has a `validateName()` function, `parseRaw` uses it internally and the new import from env uses it too. I think we should use it after the `JSON.parse` in `parseFromJSON` and throw a similar error message as the one from `parseEnvLines`.
* For the duplicate name validation, sadly, `JSON.parse` does not throw, keeps the last value and the second param `reviver` does not help. I saw a few "parsers" handling this situation but we CLEARLY WON'T use a broken specific JSON parse. I mean, who are those people? I would still like to try an idea...

IDEA: Could we compare `rawValue with `JSON.stringify(JSON.parse(rawValue))`? If those aren't strictly equal, we log a WARNING with something like `be careful, you may have duplicates in your JSON file, only the last one was used`.

I need:

1. Your opinion on name validation
2. Your opinion on my "handling duplicate properties" idea
3. Your preference about:
  a) you try to implement what I proposed
  b) I take your PR and handle it + release

hsablonniere:
Links:

* No way I'm trusting this code: https://github.com/jackyjieliu/json-dup-key-validator
* `JSON.parse` and the `reviver` param: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Using_the_reviver_parameter

hsablonniere:
OK, sorry but my idea is a bad idea, at least it's way too naïve.

```
> jsonn = '{"foo":42, "bar":36, "foo":41}'
'{"foo":42, "bar":36, "foo":41}'
> jsonn === JSON.stringify(JSON.parse(jsonn))
false
```

```
> jsonn = '{"foo":42, "bar":36, "fooo":41}'
'{"foo":42, "bar":36, "fooo":41}'
> jsonn === JSON.stringify(JSON.parse(jsonn))
false
```

Of course, because the stringify is spaceless.

hsablonniere:
I would say, let's just handle name validation and let users be careful about their JSON inputs.

divarvel:
Okay, so:
 - check that the parsed value is an object
 - use `Object.keys` to retrieve the keys and run a validation function on each key

Is there a function from `clever-client` that I can use to check variable names?

hsablonniere:
I think you missed some of my messages...

> For the name validation, @clevercloud/client/cjs/utils/env-vars.js has a validateName()

I also asked about "do you want me to handle it or can you update your PR"?

Re-reading my first comments, and thinking about what I tried with duplicated names, a `[{"name": "THE_NAME", "value": "THE VALUE"}]` JSON format would be better I think.

* It would allow us to detect duplicates properly
* It would allow us to reuse this format for the export as JSON

I need @clementd-fretlink and @celine-m-s to confirm such a JSON format is OK to produce for you?

clementd-fretlink:
ah, sorry, only the last one was displayed indeed.

I'm working on it, so I can go from there.

A list of objects would be _possible_ as well, but I'm not sure why it would be better.
Both the CC API and node JS use records to describe env values. Reading your first messages, the list format seems tied to the `KEY=value` input format, which we don't use here.

clementd-fretlink:
I've implemented a first try, which checks that the input JSON is an object, and that the keys are valid names.

Regarding the list format: if you think that simplifies the implementation by allowing to reuse code made for the `KEY=value` format, then that's a good reason, even though I think it would make things a bit more complex down the line (both for import and export)

hsablonniere:
OK, I'll try again :wink:

Here are the reasons why I prefer an array of objects with `name` and `value` properties (referred to as array format) over an object with properties being variable names (referred to as object format).

## Duplicated names

There is no proper way to parse a JSON string representing an object and detect duplicated properties in that object.

* If we use the object format, we have no way to prevent users from doing mistakes. If the user listed the same property twice, the import command will not fail nor warn and it will only keep the last value without any errors/warnings.
* If we use the array format, we'll be able to produce such errors/warnings and help our users.

## The future export as JSON command

I the near future, we'll work on a command to export env vars as JSON. For this export, we will use an array format.

* It will allow us to add more than just a `name` and `value`.
* Since the `clever env` command outputs env vars for many sources with potential duplicated names, we may add a `source` property to differentiate them.
* The object format would prevent us from being able to produce JSON will all variables for `clever env` like we do with the other format. (ex: you linked 2 PostgreSQL DB, or have a conflict between a user defined vars and vars coming from the plateform or add-ons).
* We also may annotate env vars that are used by CC (ex: `CC_JAVA_VERSION` would get a `isClever` boolean or something).

## Future JSON mode in the GUI

We're thinking about adding a JSON mode in the GUI. This means you would have 3 "tabs": SIMPLE, EXPERT and JSON.

The component maintains an internal state which is used when you update the variables. We would like to use the same JSON format in this GUI component as in this import command.

* As we already described, we won't be able to warn users about duplicated properties in this GUI component.
* There's also something not that easy to explain regarding switching between modes and maintaining a state. We need to be able to read the format but also to write it if the user changes something in one mode and switches to another one before saving.

## Security concerns

It's more of a precaution than a real problem that could cause a vuln but we tend to try to avoid using JavaSript objects with properties being strings that came from user inputs.

Here's some reading about that:

* https://hueniverse.com/a-tale-of-prototype-poisoning-2610fa170061
* https://hueniverse.com/square-brackets-are-the-enemy-ff5b9fd8a3e8

## General

* The object format is not extensible, the array format is.
* The array format has already been considered as the new format for the API, it's just too big of a change to do it right now.

Anyway, in the end, it's not really a matter of the NAME=VALUE you seem not to like :stuck_out_tongue:. It's also not a matter of reusing code we wrote for this format. It's a matter of proper validation to help the user, of extensibility (and simplification for the future GUI code). The security is a bonus.

divarvel:
Thanks for the detailed response! Consistency and extensibility are indeed two great reasons that offset the drawbacks. I'll update the PR accordingly

hsablonniere:
@divarvel You helped me to notice a bug/limitation in the GUI component.

divarvel:
:muscle: it's always good to write things down (and i'm excited for the improvement to environment variables, it will allow to unlock cool features)

hsablonniere:
FYI, the fresh GUI issue I was talking about if you're interested: https://github.com/CleverCloud/clever-components/issues/94

hsablonniere:
Thanks for you work @divarvel, I'll take it from here and refine the details ;-)

divarvel:
sure! thanks for your patience

hsablonniere:
* I reworked the validation
* I tried to detail the error messages (we'll try to use them in the future GUI)
* I also added the feature on `clever published-config`

https://github.com/CleverCloud/clever-tools/tree/import-env-json

divarvel:
Ah great, I forgot about published-config 👍

I see that it now forces all values to be strings, that's nice (but a bit
stricter that the API, which allows scalars, iirc)

The error messages are clearer 👌 (during my tests the caught JSON parsing
error was not displayed though)

hsablonniere:
:laughing: I was so focused on making sure we have an array objects with strings that I lost my "main" goal => errors on duplicates.

hsablonniere:
> I see that it now forces all values to be strings, that's nice (but a bit
stricter that the API, which allows scalars, iirc)

I was not aware of that but:

* We don't use this "feature" in the GUI or the CLI right now.
* I want to be strict here because the API will eventually cast those to a string and if I complain now, I can explain it to the user and prevent suprises.
* In the GUI (and the CLI to one day), I will need to produce a JSON out of the current env state, I won't be able to produce anything else than strings.

I just tried a few `PUT` API then some GET:

* `{"FOO": -0}` => `FOO="0"`
* `{"FOO": 1e6}` => `FOO="1000000"`

Those are actually what I would expect but I prefer to have less "magic" and less guesses.

clementd-fretlink:
I think that's better to force everything to be strings indeed, less room
for surprising behavior.

hsablonniere:
Branch should be ok right now, I'll merge and release tomorrow.

hsablonniere:
Thanks again @clementd-fretlink for your work, closing this... please see https://github.com/CleverCloud/clever-tools/pull/409

PR #332, opened 2019-07-23
Fix bad usage of colors/safe in notification

hsablonniere:


hsablonniere:
Self merge FTW :muscle:

Issue #331, opened 2019-07-23
Set build instance size


divarvel:
It could be done within `clever scale`. I can provide a patch for this change.

urcadox:
We should probably create a new command like `settings` or something at some point, but this option does make sense in the `scale` command IMO.

@hsablonniere WDYT?

hsablonniere:
Isn't it related to this https://github.com/CleverCloud/clever-tools/issues/74 ?

hsablonniere:
I'm also considering `clever scale`, I'll think about it.

divarvel:
Oh right, I forgot this issue (build flavor was not editable back then, though).

hsablonniere:
Indeed.

clementd-fretlink:
I'll start working on the feature itself, choosing the associated command can be done later

hsablonniere:
Fixed in latest beta. I also added the information in `clever status`;

PR #336, opened 2019-07-29
Add the ability to set dedicated build instances flavors

clementd-fretlink:
As discussed in #331, here's is a candidate implementation. Since we weren't sure on the UI part, I've split the change in two commits:

- one for the model part
- one for the UI part

Regardless of the UI, the model part _both_:
- toggles the `dedicatedBuildInstances` setting
- sets `buildFlavor`

This reflects the console UI where the two settings are somehow tied. It could be done with two orthogonal flags, but I fear it would be counter intuitive to be able to set a build flavor while keeping dedicated build instances disabled.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -225,6 +225,11 @@ function run () {
         return cliparse.autocomplete.words(Application('listAvailableFlavors')());
       },
     }),
+    buildFlavor: cliparse.option('build-flavor', {
+      metavar: 'buildflavor',
+      parser: Parsers.buildFlavor,
+      description: 'The size of the build instance, or `disabled` if you want to disable dedicated build instances'
+    }),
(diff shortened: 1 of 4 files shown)


hsablonniere:
@clementd-fretlink Thanks very much for this. As you may have seen in `master`, we try to promisify and "un-bacon" our code each time we modify it.

Is it OK if I rewrite your commits and apply the modifications?

divarvel:
Sure, I'm currently AFK, so I can't do it myself right now.

Thanks!

hsablonniere:
No pb, I'll handle it, there are a few rules to follow. I'll try to ship it with the upcoming 1.5.0.

hsablonniere:
I integrated your commits on master ;-)

divarvel:
Thanks!

PR #356, opened 2019-11-15
enable drains to addons using --addon (on list/create)

KannarFr:


src/commands/drain.js
```diff
@@ -8,9 +8,16 @@ const { getDrains, createDrain, deleteDrain, updateDrainState } = require('@clev
...
 async function list (params) {
-  const { alias } = params.options;
+  const { alias, addon } = params.options;
 
-  const { app_id: appId } = await AppConfig.getAppData(alias).toPromise();
+  let appOrAddonId = null;
+  if (addon) {
+    appOrAddonId = addon;
+  } else {
+    appOrAddonId = await AppConfig.getAppData(alias).toPromise();
```
KannarFr: I'm not satisfied here: if there is no addonId and application alias, the error is 

```
[ERROR] Several applications are linked. You can specify one with the "--alias" option. Run "clever applications" to list linked applications. Available aliases: prod, par0, preprod
```

I think we need custom message when addon option is available to show that addon option exists.

WDYT? @hsablonniere

Issue #357, opened 2019-11-19
Recursive parent lookup of `.git` breaks when there is no `.git`


hsablonniere:

hsablonniere:
Here's the output of the error:

```
clever deploy -v 
[DEBUG] Load configuration from environment variables
[DEBUG] Load configuration from /Users/jdoe/.config/clever-cloud
[DEBUG] Loading app configuration from /Users/jdoe/my-app/.clever.json
[STACKTRACE]
[Error: ENAMETOOLONG: name too long, stat '../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../.git'] {
  errno: -63,
  code: 'ENAMETOOLONG',
  syscall: 'stat',
  path: '../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../../.git'
}
[/STACKTRACE]
[...]

PR #363, opened 2020-02-03
solved #357 handle error when the .git folder is not found

Sacramentix:


hsablonniere:
@Sacramentix Thanks very much for this fix :wink: :clap:

ldoguin:
@Sacramentix thank you! please let us know where we can send you stickers :)

Issue #378, opened 2020-03-24
clever scale --build-flavor disabled: you provided invalid application data


BlackYoup:
The help of `clever scale --help` shows that to disable the build instance, you have to pass `disabled`: `clever scare --build-flavor disabled`

But when I do this, it throws an error:
```
~ $ clever scale --build-flavor disabled
[ERROR] You provided invalid application data
```

divarvel:
There's an issue in the request that's sent to the API. The help text is accurate

hsablonniere:
@divarvel Thanks to your feedback and PR, we investigated and managed to simplify the code and do everything in one call. See PR #380.

PR #379, opened 2020-03-24
Fix `setDedicatedBuildInstance`

clementd-fretlink:
It only sends the needed fields 

See #378

I'm not sure this is the best solution (I don't know the new clever-client well, maybe it provides a better way to do that?), but it works well.

### DIFF
--- a/src/models/application.js
+++ b/src/models/application.js
@@ -223,7 +223,16 @@ async function setScalability (appId, orgaId, scalabilityParameters) {
 
 async function setDedicatedBuildInstance (appId, orgaId, enableSeparateBuild) {
   const app = await application.get({ id: orgaId, appId }).then(sendToApi);
-  const newConfig = { ...app, separateBuild: enableSeparateBuild };
+  const newConfig = {
+    id: app.id,
+    name: app.name,
+    description: app.description,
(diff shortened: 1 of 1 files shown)


hsablonniere:
Thanks very much @clementd-fretlink, seems like we were sending way too much stuffs. I'll just check that we're not missing anything...

clementd-fretlink:
I inspected the query sent by the console to get the list of needed fields. Maybe there's a better way?

hsablonniere:
I just checked with @Keruspe and the API is actually a patch so:

* We don't need to send other data
* We can update `separateBuild` and `buildFlavor` in one call

This "bug" was introduced when we added `buildFlavor` in the payload of `GET /.../applications/{appId}`.

I'm working on a fix, this will be released today-ish...

clementd-fretlink:
ah good. Is there something I can do on this PR?

hsablonniere:
@divarvel Thanks very much for the PR, sorry we don't include it as is but as you can see in #380, after investigations on the API, we were able to simplify the code and the number of API calls.

clementd-fretlink:
Ah that's even better :+1:

PR #385, opened 2020-03-26
Add clever tcp-redirs command

Keruspe:
This requires a new version of the clever-client

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -107,6 +107,10 @@ function run () {
       parser: Parsers.commaSeparated,
     }),
     envVariableValue: cliparse.argument('variable-value', { description: 'Value of the environment variable' }),
+    port: cliparse.argument('port', {
+      description: 'port identifying the TCP redirection',
+      parser: Parsers.integer,
+    }),
   };
 
   // OPTIONS
@@ -166,6 +170,11 @@ function run () {
     deploymentId: cliparse.option('deployment-id', {
       metavar: 'deployment_id',
       description: 'Fetch logs for a given deployment',
+    }),
+    namespace: cliparse.option('namespace', {
+      metavar: 'namespace',
+      description: 'namespace in which the TCP redirection should be',
+      required: true,
     }),
     notificationEventType: cliparse.option('event', {
       metavar: 'type',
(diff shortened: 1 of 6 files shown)


bin/clever.js
```diff
@@ -681,6 +692,27 @@ function run () {
     options: [opts.alias],
   }, stop);
 
+  // TCP-REDIRS COMMAND
+  const tcpRedirs = lazyRequirePromiseModule('../src/commands/tcp-redirs.js');
+  const tcpRedirsListNamespacesCommand = cliparse.command('list-namespaces', {
+    description: 'List the namespaces in which you can create new tcp redirections',
```
hsablonniere: TCP should be in upper case ;-)

bin/clever.js
```diff
@@ -681,6 +692,27 @@ function run () {
...
+    description: 'Add a new TCP redirection to the application',
+    options: [opts.alias, opts.namespace],
+  }, tcpRedirs('add'));
+  const tcpRedirsRemoveCommand = cliparse.command('remove', {
+    description: 'Remove a TCP redirection from the application',
+    options: [opts.alias, opts.namespace],
+    args: [args.port],
+  }, tcpRedirs('remove'));
+  const tcpRedirsCommands = cliparse.command('tcp-redirs', {
+    description: 'Control the tcp redirections from reverse proxies to your application',
```
hsablonniere: TCP should be in upper case ;-)

src/commands/tcp-redirs.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+
+async function list (params) {
+  const { alias } = params.options;
+  const { org_id, app_id: appId } = await AppConfig.getAppData(alias).toPromise();
+
+  const redirs = await application.getTcpRedirs({ id: org_id, appId }).then(sendToApi);
+
+  if (redirs.length == 0) {
```
hsablonniere: Please use a `===`

hsablonniere:
@Keruspe I had a quick look, seems nice. I saw that you tried to apply all the explicit and implicit conventions of the projet ;-) Thanks :clap:

I added a few comments but there are other lint problems. Run:

```
npm run lint
```

Fix those and you should be fine :wink: 

FYI, there is also this command which tries to fix most of the lint pb:

```
npm run lint:fix
```

Keruspe:
Fixed the linting errors

Keruspe:
updated clever-client

src/commands/tcp-redirs.js
```diff
@@ -29,11 +32,31 @@ async function list (params) {
...
+
 async function add (params) {
-  const { alias, namespace } = params.options;
+  const { alias, namespace, yes: skipConfirmation } = params.options;
   const { ownerId, appId } = await AppConfig.getAppDetails({ alias });
 
-  const { port } = await application.addTcpRedir({ id: ownerId, appId }, { namespace }).then(sendToApi);
+  const { port } = await application.addTcpRedir({ id: ownerId, appId }, { namespace }).then(sendToApi).catch((error) => {
+    if (error.status === 402) {
+      return acceptPayment(error.response.body, skipConfirmation).then(() => application.addTcpRedir({ id: ownerId, appId, payment: 'accepted' }, { namespace })).then(sendToApi);
```
hsablonniere: This line is a bit long IMO.

What about:

```js
if (error.status === 402) {
  return acceptPayment(error.response.body, skipConfirmation).then(() => {
    application.addTcpRedir({ id: ownerId, appId, payment: 'accepted' }, { namespace })).then(sendToApi);
  }
}
```

I think a try/catch would be even more readable but that's just me. The then/catch is fine.

hsablonniere: Also, it helped me realize, in your code, the `sendToApi` is chained on the return of `acceptPayment` and note `application.addTcpRedir`. In this case, it changes nothing but it may introduce confusion.

hsablonniere:
* One small change (see review comment)
* A rebase on 2.2.2

and it's gooo !!! :tada:

src/commands/tcp-redirs.js
Keruspe: Fixed

Keruspe:
Rebased

PR #396, opened 2020-03-30
fix clever link

Keruspe:
Requires a new clever-client release


src/models/application.js
```diff
@@ -120,6 +120,11 @@ function get (ownerId, appId) {
   return application.get({ id: ownerId, appId }).then(sendToApi);
 };
 
+function getFromSelf (appId) {
```
hsablonniere: Can you add a comment to explicit why we need to do this?

Keruspe:
Rebased

PR #412, opened 2020-05-04
introduce clever config subcommand

### LINKED ISSUE #74: Edit application config with `clever config`
For now: name, description, zero-downtime deployment, sticky sessions, dedicated build instances, cancel on push.

`clever config set name <name>`
`clever config set description <description>`
`clever config set zero-downtime (true|false)`
`clever config set sticky-sessions (true|false)`
`clever config set dedicated-build (false|<flavor>)`
`clever config set cancel-on-push (true|false)`

It could also be done with flags, but that could get messy. Thoughts?

Keruspe:
Fixes #74

### DIFF
--- a/src/commands/config.js
+++ b/src/commands/config.js
@@ -4,75 +4,9 @@ const application = require('@clevercloud/client/cjs/api/application.js');
 
 const AppConfig = require('../models/app_configuration.js');
 const Application = require('../models/application.js');
-const Logger = require('../logger.js');
+const ApplicationConfiguration = require('../models/application_configuration.js');
 
 const { sendToApi } = require('../models/send-to-api.js');
-
-const CONFIG_KEYS = [
-  { id: 'name', name: 'name', displayName: 'Name' },
-  { id: 'description', name: 'description', displayName: 'Description' },
-  { id: 'zero-downtime', name: 'homogeneous', displayName: 'Zero-downtime deployment', displayer: zeroDowntimeDisplayer, parser: zeroDowntimeParser },
-  { id: 'sticky-sessions', name: 'stickySessions', displayName: 'Sticky sessions', displayer: booleanDisplayer, parser: booleanParser },
-  { id: 'cancel-on-push', name: 'cancelOnPush', displayName: 'Cancel current deployment on push', displayer: booleanDisplayer, parser: booleanParser },
-];
-
-function listAvailableConfigurations () {
-  return CONFIG_KEYS.map((config) => config.id);
-}
-
-function getConfigById (id) {
-  const config = CONFIG_KEYS.find((config) => config.id === id);
-  if (config == null) {
-    Logger.error(`Invalid configuration name: ${id}.`);
-    Logger.error(`Available configuration names are: ${listAvailableConfigurations().join(', ')}.`);
-  }
-  return config;
-}
-
-function zeroDowntimeDisplayer (homogeneous) {
-  return booleanDisplayer(!homogeneous);
-}
-
-function booleanDisplayer (value) {
-  return (value) ? 'enabled' : 'disabled';
-}
-
-function defaultDisplayer (value) {
-  return `${value}`;
-}
-
-function zeroDowntimeParser (value) {
-  return !booleanParser(value);
-}
-
-function booleanParser (value) {
-  return (value !== 'false');
-}
-
-function identity (value) {
-  return value;
-}
-
-function printConfig (app, config) {
-  if (app[config.name] != null) {
-    const displayer = config.displayer || defaultDisplayer;
-    Logger.println(`${config.displayName}: ${displayer(app[config.name])}`);
-  }
-}
-
-function printConfiguration (app, configurationName) {
-  if (configurationName == null) {
-    for (const config of CONFIG_KEYS) {
-      printConfig(app, config);
-    }
-  }
-  else {
-    const config = getConfigById(configurationName);
-    if (config != null) {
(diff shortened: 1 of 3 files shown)


Keruspe:
For now, it's read-only. `clever config` gets "all" the registered configuration, `clever config get foo` gets the configuration with id 'foo'.

At this stage, consider this as a RFC before going further :)

Keruspe:
ATM the set command has two arguments: the config name and the config value.

Subcommands could be generated instead from the CONFIG_KEYS if we want. That could make for better --help

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+const CONFIG_KEYS = [
+  { id: 'name', name: 'name', displayName: 'Name' },
+  { id: 'description', name: 'description', displayName: 'Description' },
+  { id: 'zero-downtime', name: 'homogeneous', displayName: 'Zero-downtime deployment', displayer: zeroDowntimeDisplayer },
+  { id: 'sticky-sessions', name: 'stickySessions', displayName: 'Sticky sessions', displayer: booleanDisplayer },
+  { id: 'cancel-on-push', name: 'cancelOnPush', displayName: 'Cancel current deployment on push', displayer: booleanDisplayer },
+];
+
+function getConfigById (id) {
+  const [config] = CONFIG_KEYS.filter((config) => config.id === id);
```
hsablonniere:
```suggestion
  const config = CONFIG_KEYS.find((config) => config.id === id);
```

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+  if (config === undefined) {
```
hsablonniere:
```suggestion
  if (config == null) {
```

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+
+function printConfig (app, config) {
+  if (app[config.name] !== undefined) {
+    const displayer = config.displayer || defaultDisplayer;
+    Logger.println(`${config.displayName}: ${displayer(app[config.name])}`);
+  }
+}
+
+function printConfiguration (app, configurationName) {
+  if (configurationName === undefined) {
```
hsablonniere:
```suggestion
  if (configurationName == null) {
```

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+function booleanDisplayer (value) {
+  return (value) ? 'enabled' : 'disabled';
+}
+
+function defaultDisplayer (value) {
+  return `${value}`;
+}
+
+function printConfig (app, config) {
+  if (app[config.name] !== undefined) {
```
hsablonniere:
```suggestion
  if (app[config.name] != null) {
```

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+
+function printConfiguration (app, configurationName) {
+  if (configurationName === undefined) {
+    for (const config of CONFIG_KEYS) {
+      printConfig(app, config);
+    }
+  }
+  else {
+    const config = getConfigById(configurationName);
+    if (config !== undefined) {
```
hsablonniere:
```suggestion
    if (config != null) {
```

src/commands/config.js
```diff
@@ -0,0 +1,66 @@
...
+function zeroDowntimeDisplayer (homogeneous) {
+  return booleanDisplayer(!homogeneous);
+}
+
+function booleanDisplayer (value) {
+  return (value) ? 'enabled' : 'disabled';
+}
+
+function defaultDisplayer (value) {
+  return `${value}`;
```
hsablonniere: Is it written this way to cast it to a string?

src/commands/config.js
```diff
@@ -63,4 +79,18 @@ async function get (params) {
...
 }
 
-module.exports = { get };
+async function set (params) {
+  const [configurationName, configurationValue] = params.args;
+  const { alias } = params.options;
+  const { ownerId, appId } = await AppConfig.getAppDetails({ alias });
+  const config = getConfigById(configurationName);
+
+  if (config !== undefined) {
```
hsablonniere:
```suggestion
  if (config != null) {
```

src/models/application_configuration.js
```diff
@@ -51,6 +53,63 @@ function parse (config, value) {
   }
 }
 
+function getUpdateOptions () {
+  return CONFIG_KEYS.flatMap((config) => getConfigOptions(config));
+}
+
+function getConfigOptions (config) {
+  switch (config.kind) {
+    case 'bool': case 'inverted-bool': {
```
hsablonniere:
```suggestion
    case 'bool':
    case 'inverted-bool': {
```

src/commands/config.js
Keruspe: Sort of. At first in some other version of the code it was to have different "symmetrical" branches, and I figured it doesn't hurt, we we can just use identity here in fact

src/models/application_configuration.js
```diff
@@ -47,6 +51,9 @@ function parse (config, value) {
     case 'inverted-bool': {
       return (value === 'false');
     }
+    case 'force-https': {
+      return (value === false) ? 'DISABLED' : 'ENABLED';
```
hsablonniere, line 55:
```suggestion
      return (value === 'false') ? 'DISABLED' : 'ENABLED';
```

Here's what I get in my test:

```
clever config set force-https false
Force redirection of HTTP to HTTPS: enabled
```

PR #413, opened 2020-05-13
Switch several things to promises

Keruspe:
Those were "trivial" to port to promises.

With this, only logs, events and activity are still using Bacon.

I tried to port the restart command but I only get the "Deployment successful" printed, not sure how to make the logs printed without porting the log models itself to promises or such, given toPromise() on the Bacon.bus seems to only give the last value


Keruspe:
```
keruspe@Lou ~/Clever/node/clever-tools (git)-[promise] % rg -l bacon src 
src/models/log.js
```

hsablonniere:
FYI, travis failed because of a 429 on project clone.

PR #418, opened 2020-05-29
Clarify `config set` / `config update` description

clementd-fretlink:
Running `clever config --help` displays the following message:

```
Available Commands:
get                        Get the current configuration
set                        Set the configuration
update                     Update the configuration
```

Here it's not clear what's the difference between `set` and `update`.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -421,19 +421,19 @@ function run () {
   // CONFIG COMMAND
   const config = lazyRequirePromiseModule('../src/commands/config.js');
   const configGetCommand = cliparse.command('get', {
-    description: 'Get the current configuration',
+    description: 'Display the current configuration',
     args: [args.configurationName],
   }, config('get'));
   const configSetCommand = cliparse.command('set', {
-    description: 'Set the configuration',
+    description: 'Edit one configuration setting',
     args: [args.configurationName, args.configurationValue],
   }, config('set'));
   const configUpdateCommand = cliparse.command('update', {
-    description: 'Update the configuration',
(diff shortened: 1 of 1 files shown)


bin/clever.js
```diff
@@ -421,19 +421,19 @@ function run () {
...
     args: [args.configurationName],
   }, config('get'));
   const configSetCommand = cliparse.command('set', {
-    description: 'Set the configuration',
+    description: 'Edit one configuration setting',
     args: [args.configurationName, args.configurationValue],
   }, config('set'));
   const configUpdateCommand = cliparse.command('update', {
-    description: 'Update the configuration',
+    description: 'Edit multiple configuration settings at the same time',
```
urcadox, line 432: "at once" sounds better to me than "at the same time" in that context

hsablonniere: @divarvel Can you modify your PR with "at once"?
I'm about to to a release for a fix on `clever restart`, I'd love to merge this at the same time.

clementd-fretlink, line 432:
```suggestion
    description: 'Edit multiple configuration settings at once',
```

clementd-fretlink: I mosty use "at once" to mean "immediately", but it works as well here :+1:

clementd-fretlink: Since this was a pure "web UI" PR, I've created two commits, a squash-merge may be in order.

hsablonniere: I was about to "Y U NO SQUASH" :laughing:

PR #420, opened 2020-06-05
Update activity.js

tolbon:
clearLine take 1 arg. I check on nodejs source undefined param become 0

in https://github.com/nodejs/node/blob/9de08f773ec4efda54ca8c9e3bc4041bade7f46f/lib/readline.js#L1257


hsablonniere:
Thanks @tolbon :tada:

PR #421, opened 2020-06-19
add domain favourite commands

jtanguy:
I had the need to display the favourite domain for an application. I figured out that setting/unsetting might be useful as well.

As for the naming: 

```
clever domain favourite get 
```
vs
```
clever domain get-favourite
```

I chose to add a `favourite` subcommand as it keeps to the same pattern of short names, I don't know if it feels too much levels of subcommands.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -477,10 +477,30 @@ function run () {
     description: 'Remove a domain name from a Clever Cloud application',
     args: [args.fqdn],
   }, domain('rm'));
+  const domainGetFavouriteCommand = cliparse.command('get', {
+    description: 'Get the favourite domain for a Clever Cloud application',
+    options: [opts.alias],
+    args: [],
+  }, domain('getFavourite'));
+  const domainSetFavouriteCommand = cliparse.command('set', {
+    description: 'Set the favourite domain for a Clever Cloud application',
+    options: [opts.alias],
+    args: [args.fqdn],
+  }, domain('setFavourite'));
+  const domainUnsetFavouriteCommand = cliparse.command('unset', {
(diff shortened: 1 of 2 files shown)


hsablonniere:
Hello Julien, I'm looking into it ;-)

hsablonniere:
@jtanguy Have you tested it? Looking at the clever client functions, it seems like you need to pass the domain in the body with `{"fqdn": "example.com"}` and not as a param with `domain` like you did.

hsablonniere:
I admit that this can be confusing compared to the add/rm domains functions.

hsablonniere:
Also unmark favourite does not need any params...

hsablonniere:
OK, I did some cleanup...

* I don't think we need the `get` subcommand
* I removed the duplicated alias options (common mistake I often make)
* I removed the param for unset
* I removed the encode for set and pass it as the body
* I added a `* ` before the favourite domain in the list

In the end, we now have:

* `clever domain` with a star prefix before the favourite domain
* `clever domain favourite` that just displays the favourite domain (no prefix)
* `clever domain favourite set example.com` to set the favourite domain
* `clever domain favourite unset` to unset the favourite domain

hsablonniere:
Merged manually with the different fixes mentioned above.

hsablonniere:
@jtanguy Thanks a lot :clap:

hsablonniere:
Released in https://github.com/CleverCloud/clever-tools/releases/tag/2.7.0 https://github.com/CleverCloud/clever-tools/blob/master/CHANGELOG.md#270-2020-08-20

PR #440, opened 2020-11-03
Access logs: handle possibly undefined source object

### LINKED ISSUE #399: [ERROR] Cannot read property 'ct' of undefined
We don't always have the city in the access logs.

BlackYoup:
Fix #399

After this fix, if the source object isn't defined, the log will appear as:
```
2020-11-02T16:31:36.512Z [ 10.2.204.148 ] GET console.clever-cloud.com /ping 304
```

I removed the localization part entirely because the `-` symbol is used as a delimiter in the extended format, not as a missing data symbol like it is in the CLF format.

I don't know if the extended log is meant to be parsed. If it is, how can we replace the source field? Using a `unknown` placeholder text? Or `missing`?

### DIFF
--- a/src/models/accesslogs.js
+++ b/src/models/accesslogs.js
@@ -19,12 +19,21 @@ function getFormatter (format, isAddon) {
   }
 }
 
+function formatSource (l) {
+  if (l.s != null) {
+    const location = l.s.ct ? `${l.s.ct}, ${l.s.co}` : l.s.co;
+    return `${l.ipS} - ${location}`;
+  } else {
+    return l.ipS;
+  }
+}
+
 function formatSimple (l) {
   return `${new Date(l.t).toISOString()} ${l.ipS} ${l.vb} ${l.path}`;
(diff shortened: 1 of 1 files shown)


src/models/accesslogs.js
```diff
@@ -19,12 +19,21 @@ function getFormatter (format, isAddon) {
   }
 }
 
+function formatSource (l) {
+  if (l.s) {
```
hsablonniere, line 23:
```js
  if (l.s != null)
```

hsablonniere (changes requested):

hsablonniere:
@BlackYoup I don't think it's meant to be parsed. I'm ok with just removing it like you did.

src/models/accesslogs.js
BlackYoup: updated

hsablonniere:
@BlackYoup Thanks :wink:

PR #443, opened 2020-11-26
deploy: for git add remote to prevent race conditions on mono repos

hsablonniere:


hsablonniere:
Needs to be polished but it has fixed the problem for the user (tested through the beta).

PR #444, opened 2020-12-17
format .clever.json for readability

mickaelchanrion:
The generated file `.clever.json` could be formatted with indentations for readability convenience.


hsablonniere:
Nice idea, thanks... We'll release this in january.

hsablonniere:
Did I say January? #sorry

Issue #449, opened 2021-01-18
Add an option to create an add-on with an encrypted disk


BlackYoup:
Add-ons will now have the option to be created with their data disk. Those add-ons are:
- PostgreSQL
- MySQL
- Redis
- MongoDB
- Jenkins
- Elasticsearch

The encryption feature is currently free but will be billed 10% of the add-on's plan in a few months (no real ETA). 

Only dedicated add-ons will support this feature, DEV or Free plans don't support this feature.

To list the add-ons options, we can use the `api.clever-cloud.com/v4/addon-providers/<provider-id>` endpoint, currently only available in this PR: https://github.com/CleverCloud/clever-client.js/pull/36. We might already be using the `api.clever-cloud.com/v2/products/addonprovider/<provider-id>/versions` endpoint to get the versions of the dedicated and shared add-on but this can be dropped because `api.clever-cloud.com/v4/addon-providers/<provider-id>` also lists versions (and features specific to each version).

To create the add-on with the `encryption` feature enabled, you have to put it in the `options` field of the body used in the create request. This should be:
```javascript
const options = {
  "encryption": "true"
};
```

Note the `"true"` which is a string and not a boolean. This is needed because the `options` object is a `HashMap<String, String>` object for now. If the `encryption` feature is disabled, it doesn't have to be in the `options` object or can be set to `"false"`.

PR #452, opened 2021-02-01
Add-on: list versions and options of a specific provider, create add-on with available options

BlackYoup:
This PR does two things that are a bit linked, I could split it in two, you tell me your preferences. This also updates the clever-client to the 7.x branch.

This is linked to #449 .

First, the `clever addon providers show <provider>` will now display available versions and features for each plan, when available. An example with Elasticsearch that has multiple options available

```
Plan xs                  
  CPUs: 1                    
  DISK: 10 GB                
  MEMORY: 1 GB                      
  NODES: 1              
  Type: Dedicated            
  Available versions: 6, 7 (default)
  Options for version 6:     
    kibana: default=false
    apm: default=false              
    encryption: default=false
  Options for version 7:     
    kibana: default=false           
    apm: default=false       
    encryption: default=false
```

Second, the `clever addon create` can now take two new arguments:
- `--addon-version` which can set the version of the add-on you want to create. I could have named it `--version` but it prints clever-tools's version. I didn't really look into how to fix that.
- `--option feature=state` (`--option encryption=true`) to add options when creating your add-on.

For the `--option`, it takes a value of the form `feature=state` because we might want to have options enabled by default at some point. This gives us the possibility to use `--option encryption=false` to disabled it if we ever enable it by default. The API already supports options enabled by default. The `--option` can take multiple values:
- true
- false
- enabled
- disabled

For example, to create a PostgreSQL add-on on version 11 with the encryption at rest enabled on the smaller plan:
```
clever addon create postgresql-addon test-from-cli --plan XXS_SML --addon-version 11 --option encryption=enabled
```

Let me know what you think of this.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -297,6 +297,14 @@ function run () {
       description: 'Region to provision the addon in, depends on the provider',
       complete: Addon('completeRegion'),
     }),
+    addonVersion: cliparse.option('addon-version', {
+      metavar: 'addon-version',
+      description: 'The version to use for the add-on',
+    }),
+    addonOptions: cliparse.option('option', {
+      metavar: 'option',
+      description: 'Option to enable for the add-on. Multiple --option argument can be passed to enable multiple options',
+    }),
     region: cliparse.option('region', {
       aliases: ['r'],
       default: 'par',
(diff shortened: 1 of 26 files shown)


hsablonniere:
> This PR does two things that are a bit linked, I could split it in two, you tell me your preferences. This also updates the clever-client to the 7.x branch.

It's OK :wink: 

> First, the clever addon providers show <provider> will now display available versions and features for each plan, when available. An example with Elasticsearch that has multiple options available

I just tried, I'm ok with this. It could be better with a table display, it's not really important so I created an issue: https://github.com/CleverCloud/clever-tools/issues/455

> Second, the clever addon create can now take two new arguments:
> --addon-version which can set the version of the add-on you want to create. I could have named it --version but it prints clever-tools's version. I didn't really look into how to fix that.

I think it's ok like this. I don't want use to introduce confusion or to invest too much time into cliparse inner details.


> --option feature=state (--option encryption=true) to add options when creating your add-on.

I like that, but we already discussed having the same system as for app options/settings.

hsablonniere:
@BlackYoup Thanks for the clever-client update BTW.

hsablonniere:
@BlackYoup Code review: LGTM

hsablonniere:
@BlackYoup I just did some tests, LGTM :+1:

PR #460, opened 2021-02-15
Networkgroups CLI

RemiBardon:


hsablonniere:
First step of the review:

* Along with @RemiBardon we did some adjustments (see additionnal commits)
  * Most important thing was error handling with a deferred
* New commands files are OK now
* New models files are OK now

## TODO for @RemiBardon 

* move away from callback based `fs` to promise based fs
  * with `const { promises: fs } = require('fs');`
* WARNING: the option `orgaIdOrName` was renamed from `org` to `orga` but it's used by other commands
  * see 66a91c229682cffd748c1bc0ce8717822f379f72
* Lots of autocomplete are commented
* Still a few FIXME to address
* Ask colleagues if we really need two commands `ng` and `networkgroups`

RemiBardon:
Payloads encoding/decoding is broken until <https://github.com/CleverCloud/clever-client.js/issues/43> is fixed.

RemiBardon:
As I said in https://github.com/CleverCloud/clever-client.js/issues/43, I recently realised only the doc comments will be affected, `body` is passed as is. It will not break the CLI.

RemiBardon:
Just posting the TODO list again so I can check boxes:

- [ ] Move away from callback based `fs` to promise based `fs`
  - with `const { promises: fs } = require('fs');`
- [x] **WARNING:** the option `orgaIdOrName` was renamed from `org` to `orga` but it's used by other commands
  - see [`66a91c2`](https://github.com/CleverCloud/clever-tools/commit/66a91c229682cffd748c1bc0ce8717822f379f72)
- [ ] Lots of autocomplete are commented
- [ ] Still a few `FIXME`s to address
- [ ] Ask colleagues if we really need two commands ng and networkgroups

RemiBardon:
@hsablonniere Autocomplete is not working on macOS (as far as I know), how can I test it while developing? Do I write it and ask someone else to try it out for me?

RemiBardon:
I added two commands for networkgroups: `clever networkgroups` and `clever ng`. They do the exact same thing, they're "aliases".

Do we keep them both? Do we choose one? Now is time to choose, as I hope this PR will be merged soon.

RemiBardon:
It has been decided that we keep both `clever networkgroups` and `clever ng` for now.

RemiBardon:
The `clever networkgroups join` command is taking a long time to release, because the required elevation causes lots of problems.

I think it'd be better to release the `networkgroups` command without the `join` (and `leave`), allowing a user to administrate their networkgroups. `join` is a "Nice to have" command, and should not block the entire release.

@hsablonniere What do you think?

I will check the commands to see if everything is alright, and I'll tell you when it's ready 😉 I also have [a few things to do](https://github.com/CleverCloud/clever-tools/pull/460#issuecomment-867459764), but it won't be long (hopefully).

RemiBardon:
@waxzce is ok with a separated release for the `join`/`leave` commands

PR #463, opened 2021-03-30
conf: move configuration file inside a clever-cloud directory

### LINKED ISSUE #462: ~/.config/clever-cloud should be a directory and not a file
This would allow other software to write their conf file there

Keruspe:
Fixes #462


Keruspe:
I'm not a big fan of the side effect in the config init, nor with the fact that all of this is sync, but it's the path of lesser pain since it's the only common path to everything using this

Keruspe:
Wil fix the ENOENT CI failure later.

hsablonniere:
Rebased and OK ;-)

Issue #334, opened 2019-07-23
Add a clever status --all


hsablonniere:

hsablonniere:
While working on the `--all` option, I tried an idea and I think it's the right occasion to update the list of informations of `clever status`.

Here's the simple POC I tried:

![Capture d’écran_2019-09-05_18-40-09](https://user-images.githubusercontent.com/236342/64418883-165a4100-d09c-11e9-8a51-4f8c943b6413.png)

hsablonniere:
I'm proposing something like that:

```
Name              MyApp            MyApp (staging)
Alias             prod             staging
Status            running 2*M      running 1*S
//                stopped          stopped
//                deploying 2*M    deploying 1*S
Last deploy       3 days ago       1 day ago
Running commit    391d55012475     884ce0bac349
Deploying commit  391d55012475     884ce0bac349
Git remote head   391d55012475     884ce0bac349
Auto scalability  enabled          disabled
Instances         2 to 5           1
Sizes             M                S
Dedicated build   disabled         M
```

PR #482, opened 2021-07-04
Improve status display and add --all params ( #334 )

Sacramentix:
I added an optionnal --all params that display all info of all apps in a text-table ( the npm packages that already used in the project ).


hsablonniere:
Thanks a lot for the contribution! We're going to revisit this later as part of #334, so we'll close this PR for now, but everything here will be useful when we pick it back up. Feel free to follow along on the issue.

PR #509, opened 2022-02-25
Added rpm signature.

zepag:
- Added some pretty colored messages at start/end of tasks ;)
- Added some required variables for GPG sign.
- Added gpg setup and gpg sign to the scripts, as well as .rpmmacros
  file generation.
- Not using fpm's --rpm-sign, but directly rpm --add-sign as per: https://github.com/jordansissel/fpm/issues/1720

### DIFF
--- a/scripts/config.js
+++ b/scripts/config.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const os = require('os');
 const pkgJson = require('../package.json');
 const semver = require('semver');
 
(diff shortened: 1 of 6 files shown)


scripts/config.js
```diff
@@ -96,6 +97,14 @@ function getNpmToken () {
   return token;
 }
 
+function getGpgConf () {
+  const gpgPrivateKey = process.env.RPM_GPG_PRIVATE_KEY;
+  const gpgPath = process.env.RPM_GPG_PATH || os.homedir();
+  const gpgName = process.env.RPM_GPG_NAME;
+  const gpgPass = process.env.RPM_GPG_PASS;
+  return { gpgPrivateKey, gpgPath, gpgName, gpgPass}
```
Keruspe, line 105: Does it pass elint with the missing space before brace?

zepag: I didn't change the current build process, just amended it to fix the aforementioned missing rpm signature.
No idea about that, though there weren't any eslint complains in the build AFAIR, (probably because those build scripts are not linted ?).
This being said, this is just a convention (I love them, don't misunderstand me) and not a blocker.

hsablonniere:
This is awesome!!! :tada:

PR #552, opened 2023-01-24
feat(database): add command to download backups

miton18:


src/commands/database.js
```diff
@@ -0,0 +1,45 @@
...
+const { println } = require('../logger.js');
+const formatTable = require('../format-table')();
+const superagent = require('superagent');
+
+async function listBackups(params) {
+    const [ databaseId ] = params.args
+    const ownerId = params.options.org.orga_id;
+
+    const req = await getBackups({ ownerId, ref: databaseId })
+    const backups = await sendToApi(req)
```
hsablonniere, line 14: Most of the time we do this:

```js
const backups = await getBackups({ ownerId, ref: databaseId }).then(sendToApi);
```

src/commands/database.js
```diff
@@ -0,0 +1,45 @@
...
+
+    println(formatTable(formatted));
+}
+
+async function downloadBackups(params) {
+    const [ databaseId, backupId ] = params.args
+    const ownerId = params.options.org.orga_id
+
+    const req = await getBackups({ ownerId, ref: databaseId })
+    const backups = await sendToApi(req)
```
hsablonniere, line 32: Same here

hsablonniere (changes requested):
Hey @miton18, nice work!!

Some of the code formatting needs to be adjusted but you can't guess it as there's no automation for this. I'll adjust it before the merge for you.

I think you're not using the right version of Node.js and npm and this caused the `package-lock.json` to go from version 1 to 2. This caused a few problems in the past and we try to keep it like that for now. I'll handle it before the merge for you.

I put a few "change requests" and I haven't tested the code but apart from that this seems OK ;-)

hsablonniere:
* (1) I have some doubts on the content encoding of the dump, we need to try an import
  * Seems like @miton18 tried to use the output with a `pg_restore -l`
* (2) We need table headers for the list of backups, I can add it
* (3) Do we need a parameter to output the dump directly to a file?
  * Like `curl -o`
* (4) What about allowing not specifying the `orga-id` or `database-id` and figuring it out from the `/summary`?
  * This means more API calls but whatever, it's way easier for users
  * Does this mean this parameters are optional or to be removed entirely?

miton18:
> * (1) I have some doubts on the content encoding of the dump, we need to try an import
>       
>       * Seems like @miton18 tried to use the output with a `pg_restore -l`

In fact,
```
node ./bin/clever.js database backups download -o orga_xxx postgresql_yyy zzz | pg_restore -l
;
; Archive created at 2023-02-04 02:33:38 CET
;     dbname: bi4rthuir
;     TOC Entries: 183
;     Compression: -1
;     Dump Version: 1.14-0
;     Format: CUSTOM
;     Integer: 4 bytes
;     Offset: 8 bytes
;     Dumped from database version: 14.4
;     Dumped by pg_dump version: 14.4
;
;
; Selected TOC Entries:
;
7711; 0 0 ACL - SCHEMA public ufd6w7
```
 

>     * (2) We need table headers for the list of backups, I can add it

Adding

>     * (3) Do we need a parameter to output the dump directly to a file?
>       
>       * Like `curl -o`

Adding

>     * (4) What about allowing not specifying the `orga-id` or `database-id` and figuring it out from the `/summary`?
>       
>       * This means more API calls but whatever, it's way easier for users
>       * Does this mean this parameters are optional or to be removed entirely?

Yes, I can, look summary if no org ID is specified, look database list if no database ID is specified

It's easier but slower.

possible alternatives:

* today we link apps to a project, we can also link addons ?
* context switching has been addresses an other way for example in `kubectl` commands, this one use a _context file_ located in the user home dir, which tell which cluster/namespace to use. We can bring this concept in clever-tools. For exemple, in this file, we can store a map

```json
{
    "contexts": {
        "prod": { // user defined alias which is not linked to organization name
            "endpoint": "https://api.clever-cloud.com",
            "organization": "orga_xxx",
        },
        "dev": {
            "endpoint": "https://api.clever-cloud.com",
            "organization": "orga_zzz",
        }
    },
    "default": "prod"
}
```

So users can use cli without `--org` assuming it's org_xxx by default, or use an optional option

```sh
clever-tools --on dev database backups postgresql_uuu
```

hsablonniere:
Thank you @miton18 :clap: 

About the :

> today we link apps to a project, we can also link addons ?
> context switching has been addresses an other way for example in kubectl commands

I think we need to go in this kind of direction in the long term.
We could also have a local "cache" that associates an org to an addon ID or an org ID since it's immutable.

We'll discuss this in the workshop.

PR #571, opened 2023-05-15
Fix ng term

KannarFr:
image
Here is a full example of the results.

### DIFF
--- a/src/models/format-ng-table.js
+++ b/src/models/format-ng-table.js
@@ -78,8 +78,8 @@ const peersTableColumnLengths = [
   15, /* ip */
 ];
 const formatPeersTable = formatNgTable(peersTableColumnLengths);
-function formatPeersLine (peer) {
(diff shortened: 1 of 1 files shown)


RemiBardon:
Unless it changed, spaces before `()` in function declarations are enforced by the linter configuration, otherwise LGTM

KannarFr:
Where is this linter configuration? I defined no linter in my editor and it auto-applied these changes.

RemiBardon:
<https://github.com/CleverCloud/clever-tools/blob/master/.eslintrc.js> doesn't override the [`space-before-function-paren`](https://eslint.org/docs/latest/rules/space-before-function-paren) rule, which is enabled by default (<https://eslint.org/docs/latest/rules/space-before-function-paren#options>)

RemiBardon:
I guess you can just drop 9906546058425043fa3d26eb0a569dd2621e622e to fix the eslint warnings

KannarFr:
Done.

RemiBardon (approved):

hsablonniere (approved):

PR #596, opened 2023-08-22
feat(diag): add config file and auth source to clever diag

### LINKED ISSUE #594: Add some details in the `clever diag`
## Context

We encountered the following situation multiple times:

* The user sets `CLEVER_TOKEN` and `CLEVER_SECRET` in a bash/zsh profil
* The user forgets about this
* The user tried to `clever login` but the CLI still uses the env vars
* If the token from the env vars is revoked, it appears like the clever login fails

## TODO

We should add some informations in the `clever diag` output:

* The location of the config file
* The source of the tokens (env vars or config file)

hsablonniere:
Closes #594

### DIFF
--- a/src/commands/diag.js
+++ b/src/commands/diag.js
@@ -28,6 +28,8 @@ async function diag () {
     Logger.println('Linux         ' + colors.green(linuxInfos));
   }
 
+  Logger.println('Shell         ' + colors.green(process.env.SHELL));
+
   const isPackaged = (process.pkg != null);
   Logger.println('Packaged      ' + colors.green(isPackaged));
   Logger.println('Exec path     ' + colors.green(process.execPath));
--- a/src/commands/diag.js
+++ b/src/commands/diag.js
@@ -8,12 +8,12 @@ const colors = require('colors/safe');
 const Logger = require('../logger.js');
 const pkg = require('../../package.json');
 const User = require('../models/user.js');
-const { loadOAuthConf } = require('../models/configuration.js');
+const { conf, loadOAuthConf } = require('../models/configuration.js');
 
 async function diag () {
 
   const userId = await User.getCurrentId().catch(() => 'Not connected');
-  const oauthToken = await loadOAuthConf().then((o) => o.token).catch(() => 'Not connected');
+  const authDetails = await loadOAuthConf();
 
   Logger.println('clever-tools  ' + colors.green(pkg.version));
   Logger.println('Node.js       ' + colors.green(process.version));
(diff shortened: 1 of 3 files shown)


src/commands/diag.js
```diff
@@ -28,12 +28,16 @@ async function diag () {
...
+
   const isPackaged = (process.pkg != null);
   Logger.println('Packaged      ' + colors.green(isPackaged));
   Logger.println('Exec path     ' + colors.green(process.execPath));
+  Logger.println('Config file   ' + colors.green(conf.CONFIGURATION_FILE));
 
   Logger.println('User id       ' + colors.green(userId || 'Not connected'));
-  Logger.println('oAuth token   ' + colors.green(oauthToken));
+  Logger.println('oAuth token   ' + colors.green(authDetails.token || 'Not connected'));
+  Logger.println('Auth source   ' + colors.green(authDetails.source));
```
aurrelhebert, line 40: I would have put this information above or just below the config file path

src/commands/diag.js
```diff
@@ -28,12 +28,16 @@ async function diag () {
...
   }
 
+  Logger.println('Shell         ' + colors.green(process.env.SHELL));
+
   const isPackaged = (process.pkg != null);
   Logger.println('Packaged      ' + colors.green(isPackaged));
   Logger.println('Exec path     ' + colors.green(process.execPath));
+  Logger.println('Config file   ' + colors.green(conf.CONFIGURATION_FILE));
 
   Logger.println('User id       ' + colors.green(userId || 'Not connected'));
```
aurrelhebert, line 38: In case the user is not connected, maybe we can hide the information below (the oAuthToken and the auth source)?

aurrelhebert:
LGTM only small comment :)

src/commands/diag.js
```diff
@@ -28,12 +28,16 @@ async function diag () {
...
-  Logger.println('oAuth token   ' + colors.green(oauthToken));
+  Logger.println('oAuth token   ' + colors.green(authDetails.token || 'Not connected'));
```
aurrelhebert, line 39: And maybe print `Not connected` in an other color?

src/commands/diag.js
hsablonniere: Agreed! :+1:

src/commands/diag.js
hsablonniere: Agreed! :+1:

src/commands/diag.js
hsablonniere: Agreed! :+1:

aurrelhebert (approved):

PR #598, opened 2023-08-29
feat(deploy): add same-commit-policy option

### LINKED ISSUE #276: Add an option so `clever deploy` can do a `clever restart` automatically if remote commit is the same
Actually, if you want to deploy the same commit twice, even with `-f` option, the command will return an error `[ERROR] The clever-cloud application is up-to-date. Try this command to restart the application: clever restart`.

Without `force` option, `restart` sounds like fine.
However `deploy` with `--force` option should still be OK.

Current use case is : deploying from Gitlab-CI, triggering manual action that will rebuild the website and run every CC HOOKS.

aurrelhebert:
Closes #276


hsablonniere:
LGTM, I just need to do some QA and we'll be able to merge this ;-)

miton18:
can you add a new line ?

```
``
[ERROR] The clever-cloud application is up-to-date (7a8c944350ead07842ff95162f230dbda747dd74). You can set a policy with 'same-commit-policy' to handle differently when remote HEAD has the same commit as the one to push. Or try this command to restart the application:
        clever restart
`

**Review by hsablonniere — CHANGES_REQUESTED** [review]

File: src/command-options.js [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+function getSameCommitPolicyOption () {
+  const availablePolicies = ['errors', 'ignore', 'restart', 'rebuild'];
+  return cliparse.option('same-commit-policy', {
+    aliases: ['p'],
+    metavar: 'same-commit-policy',
+    parser: (policy) => {
+      return availablePolicies.includes(policy)
+        ? cliparse.parsers.success(policy)
+        : cliparse.parsers.error('The output policy must be one of ' + availablePolicies.join(', '));
```
miton18, line 29:
```suggestion
        : cliparse.parsers.error('The output policy must be one of  ${availablePolicies.join(', ')});
```

src/command-options.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    default: 'errors',
+    description: `Which policy to apply when the same commit is the same as the remote one. Availabe policies are (${availablePolicies.join(', ')})`,
```
miton18, line 32:
```suggestion
    description: `Which policy to apply when the local commit is the same as the remote one. Available policies are (${availablePolicies.join(', ')})`,
```

miton18 (approved):

hsablonniere (approved):
Well done @aurrelhebert :+1: :clap:

PR #602, opened 2023-09-29
feat: use new logs API

miton18:


hsablonniere:
I moved the JSON code to https://github.com/CleverCloud/clever-tools/tree/feat/json-logs so we can handle this part separately.

This was not real JSON:

* no double quotes
* no array and comas between logs

and this was not backported to the add-on logs.

PR #607, opened 2023-10-19
Feat/json logs

miton18:
Support `--format` arg in logs command

To test it 
```
node ./clever.js logs -a xxx --since "2023-11-27T00:00:00" --until "2023-11-28T00:00:00" -F json
node ./clever.js logs -a xxx --since "2023-11-27T00:00:00" --until "2023-11-28T00:00:00" -F json-stream
node ./clever.js logs -a xxx --since "2023-11-27T00:00:00" --until "2023-11-28T00:00:00" -F human
node ./clever.js logs -a xxx --since "2023-11-27T00:00:00"  -F json <= Throw an Error 
```

both JSON format are compatible with `jq` command. Just append `| jq` at the end

`json-stream` follow the ndjson format

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -135,6 +135,7 @@ function run () {
     sourceableEnvVarsList: cliparse.flag('add-export', { description: 'Display sourceable env variables setting' }),
     accesslogsFormat: getOutputFormatOption(['simple', 'extended', 'clf']),
     addonEnvFormat: getOutputFormatOption(['shell']),
+    logsFormat: getOutputFormatOption(['json-stream']),
     accesslogsFollow: cliparse.flag('follow', {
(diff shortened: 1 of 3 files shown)


hsablonniere:
https://ndjson.org/

aurrelhebert (commented):
LGTM except that I still got an error when using `since` and `until`, which is the same for all formats: 
```
[ERROR] Error: retry is not enabled: TypeError: Cannot read properties of undefined (reading 'length')
```

aurrelhebert (approved):

hsablonniere (approved):
Tested with built preview:

* format json
  * with and without until
  * through `jq`, `jless`
* format json-stream
  * through `jq`
* format human

PR #609, opened 2023-10-20
feat(logs): support duration in since parameters

miton18:


aurrelhebert (approved):
LGTM and very helpful ;) What about ISODuration : https://en.wikipedia.org/wiki/ISO_8601#Durations? Which format of durations should we support?

aurrelhebert:
It looks like there is a lib to support ISO_DURATION: https://www.npmjs.com/package/iso8601-duration?

aurrelhebert (approved):

PR #626, opened 2023-11-06
ci: add build github action workflow

### LINKED ISSUE #625: Implement build workflow with Github action
We want to move from Jenkins to Github actions.

The first thing to do is to perform the build with a Github action workflow.

### to be discussed

When should the build workflow should be launched?
We have two options:
- like with Jenkins multibranch pipeline, the workflow is launched on every commit on any branch (including master).
- like clever-components, the workflow is launched  for every commit on branches that have an associated PR (and also on commit on the master branch)

pdesoyres-cc:
Fixes #625


aurrelhebert (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
+name: 'Build'
+on:
+  push:
+    branches:
+      - master
```
hsablonniere, line 5: question: with Jenkins, we used to build all branches but with this new system, I wonder if we should just build for PRs since the built status is only useful to block/allow PRs and the built artefacts are only published for PRs.

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+          cache: 'npm'
+      - name: '[Prepare] Install dependencies'
+        run: npm ci
+
+      - name: '[Code quality] Check commit message'
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
+
+#      - name: '[Code quality] Check ESLint'
+#        run: npm run lint
```
hsablonniere, line 32: question: is it commented because the lint command fails on master?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,112 @@
...
+  CC_CLEVER_TOOLS_PREVIEWS_CELLAR_KEY_ID: ${{ secrets.CC_CLEVER_TOOLS_PREVIEWS_CELLAR_KEY_ID }}
+  CC_CLEVER_TOOLS_PREVIEWS_CELLAR_SECRET_KEY: ${{ secrets.CC_CLEVER_TOOLS_PREVIEWS_CELLAR_SECRET_KEY }}
+
+jobs:
+  wait_for_build:
+    name: 'Wait for build to succeed'
+    if: |
+      (github.event.action == 'synchronize' || github.event.action == 'opened') &&
+      github.event.repository.fork == false &&
+      !startsWith(github.event.pull_request.head.ref, 'release-please--')
```
hsablonniere, line 20: nitpick: can we put the `&&` on the new line like this?

```
jobs:
  wait_for_build:
    name: 'Wait for build to succeed'
    if: |
      (github.event.action == 'synchronize' || github.event.action == 'opened')
      && github.event.repository.fork == false
      && !startsWith(github.event.pull_request.head.ref, 'release-please--')
```

.github/workflows/preview.yml
```diff
@@ -0,0 +1,112 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: `🔎 A preview has been automatically published:\n\n${links}\n\n_This preview will be deleted once this PR is closed._`
+            });
+  delete:
+    if: |
+      github.event.action == 'closed' &&
+      github.event.repository.fork == false &&
+      !startsWith(github.event.pull_request.head.ref, 'release-please--')
```
hsablonniere, line 84: nitpick: can we put the `&&` on the new line like this?

```
  delete:
    if: |
      github.event.action == 'closed'
      && github.event.repository.fork == false
      && !startsWith(github.event.pull_request.head.ref, 'release-please--')
```

RELEASE.md
```diff
@@ -2,146 +2,259 @@
...
+Or with auto fix:
+```shell
+npm run lint:fix
+```
+
+#### Build binaries
+
+This step does three things:
+
+* Builds binaries for GNU/Linux, macOS and Windows using [pkg](https://github.com/zeit/pkg).
```
hsablonniere, line 50: nitpick: not your fault as the link was like this before but the repo is https://github.com/vercel/pkg

.github/workflows/build.yml
pdesoyres-cc: Exactly

pdesoyres-cc:
Bundle or not bundle during the build phase?

- I agree that having to install `fpm` even though we don't need the result is pointless.
- In the meantime, I also like the build job to verify as many things as it can.

I have 2 propositions:

- add a parameter to the `job-build.js` to skip the bundling phase.
  - this is quick to implement
  - we could go further and add another parameter that skips the archive phase which you also don't need locally.
  - locally, you could just do something like `node scripts/job-build.js --skip-bundle --skip-archive`
- another proposition is to remove the bundling phase from the `job-build.js` but add `job-bundle.js` which only bundles.
  - locally you'll then run `node scripts/job-build.js` and it doesn't bundles.
  - we need to add the bundling phase to the build GitHub action.
  - the advantage is that we can install the dependencies needed by the bundling phase after the build, which may make the workflow fail faster in case of build failure.

I'm ok with both propositions, just let me know if you prefer one or another.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,59 @@
...
+  RPM_GPG_NAME: 'Clever Cloud Nexus (rpm)'
+  RPM_GPG_PASS: ${{ secrets.RPM_GPG_PASSPHRASE }}
+  NEXUS_USER: ${{ secrets.NEXUS_USER }}
+  NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+  NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
+  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+jobs:
+  publish_cellar:
+    name: 'Publish to Cellar'
```
miton18, line 21: I only see an upload to Github ?

RELEASE.md
miton18, line 50: https://github.com/vercel/pkg

RELEASE.md
```diff
@@ -2,146 +2,259 @@
...
+> Required environment variables: `RPM_GPG_NAME`, `RPM_GPG_PRIVATE_KEY`, `RPM_GPG_PASS`
+
+### Preview
+
+This workflow handles preview creation and deletion.
+
+* The creation is triggered when a PR is created or synchronized (a new commit occurs on the branch).
+* The deletion is triggered when a PR is closed.
+
+Previews are publish on Clever Cloud's Cellar.
```
miton18, line 75:
```suggestion
Previews are publish on [Cellar](https://www.clever-cloud.com/product/cellar-object-storage/).
```

RELEASE.md
```diff
@@ -2,146 +2,259 @@
...
+```
+
+Deleting a preview for the current branch:
+```shell
+./scripts/job-preview.js delete
+```
+
+### Release
+
+The release workflow is triggered when a commit is done on the `master` branch or on `hotfix/*` branches.
```
miton18, line 124:
```suggestion
The release workflow is triggered when a commit is pushed on the `master` branch or on `hotfix/*` branches.
```

RELEASE.md
```diff
@@ -2,146 +2,259 @@
...
+
+This workflow uses [release-please](https://github.com/googleapis/release-please) GitHub action.
+
+Basically, a GitHub Pull Request is maintained by this GitHub action.
+This Pull Request contains the modification for the next release.
+Merging the release Pull Request will perform the release:
```
miton18, line 130:
```suggestion
Merging the [release-please](https://github.com/googleapis/release-please) pull request will perform the release:
```

scripts/archive.js
```diff
@@ -0,0 +1,61 @@
+const fs = require('fs-extra');
```
miton18, line 1: Can we have a `#/usr/bin/env node` on every script ?

scripts/cellar-client.js
```diff
@@ -0,0 +1,91 @@
+const AWS = require('aws-sdk');
+const fs = require('fs-extra');
+const mime = require('mime-types');
+const { getCellarConf } = require('./config.js');
+
+mime.types['sha256'] = 'text/plain';
```
miton18, line 6: maybe a comment on why ?

miton18 (changes requested):

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    name: 'Build'
```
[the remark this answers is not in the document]
pdesoyres-cc: I don't understand. What do you mean?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,112 @@
...
+        id: check_preview
+        run: |
+          if node scripts/job-preview.js get ${BRANCH} > /dev/null 2>&1; then
+            echo "exists=yes" >> $GITHUB_OUTPUT
+          else
+            echo "exists=no" >> $GITHUB_OUTPUT
+          fi
+
+      - name: '[Run] Build preview'
+        run: node scripts/job-preview.js build ${BRANCH}
```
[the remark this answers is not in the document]
pdesoyres-cc: ${BRANCH} is an env var that was setup at the beginning of the process (line 10). This env is injected in every steps.
I prefer setting BRANCH once, because it make the workflow easier to read.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,112 @@
...
+        env:
+          BRANCH: ${{ github.event.pull_request.head.ref }}
+      - name: '[Run] Publish preview'
+        run: node scripts/job-preview.js publish ${BRANCH}
```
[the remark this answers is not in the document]
pdesoyres-cc: same as above

.github/workflows/preview.yml
```diff
@@ -0,0 +1,112 @@
...
+          if node scripts/job-preview.js get ${BRANCH} > /dev/null 2>&1; then
+            echo "exists=yes" >> $GITHUB_OUTPUT
+          else
+            echo "exists=no" >> $GITHUB_OUTPUT
+          fi
+
+      - name: '[Run] Build preview'
+        run: node scripts/job-preview.js build ${BRANCH}
+        env:
+          BRANCH: ${{ github.event.pull_request.head.ref }}
```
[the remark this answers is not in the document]
pdesoyres-cc: You're right, this is not needed. Thank you.

.github/workflows/publish.yml
pdesoyres-cc: The name of this step is wrong because it publishes on every target (see line 51).

scripts/archive.js
pdesoyres-cc: It makes sense to do that on entry points script. For instance, `archive.js` is just a bunch of exported utility functions, so executing it is useless.

So I'll do that on every `job-*.js` files. (I will also `chmod +x` them)

pdesoyres-cc:
* I rebased on master
* I plan to squash all commits. Is it ok for you? @hsablonniere

.github/workflows/build.yml
hsablonniere: OK :+1:

.github/workflows/build.yml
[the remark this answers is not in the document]
hsablonniere: @miton18 told me he didn't see the call to `pkg` so it's OK

hsablonniere:
@pdesoyres-cc 

* OK to squash all commits
* I think I prefer the proposition to split `job-build.js` and `job-bundle.js` because of the `may make the workflow fail faster in case of build failure`
  * but I'd rather merge this PR, do a first release and split the build/bundle later via a new issue

hsablonniere (approved):

Issue #615, opened 2023-11-02
List clever curl in command help


hsablonniere:

davlgd:
In the help, after the available commands, we could add (for example) : 

>>>
...
webhooks                   Manage webhooks

Need to go further ? Once logged in, you can use clever-tools to easily query our API through 'clever curl' command. For example : 

clever curl https://api.clever-cloud.com/v2/organisations
clever curl https://api.clever-cloud.com/v2/organisations/<YOUR ID>/applications | jq '.[].id'

Our API documentation is available here : https://www.clever-cloud.com/doc/openapi
Tell us more about what you need for Clever tools : <short feedback URL>
>>>

davlgd:
Note that `clever curl --help` prints an bad help message in french : 

```
clever curl --help
Options :
  --help     Affiche l'aide                                            [booléen]
  --version  Affiche le numéro de version                              [booléen]
```

If you type `clever curl --version` the answer is the clever-tools version number

hsablonniere:
@davlgd We like your of showing examples for clever curl but only with `clever curl --help`.

FYI: right now, we have no idea why there's some french prompt with `clever curl --help`.

chmuche:
It's possible to shortcut the URL too?

`clever curl https://api.clever-cloud.com/v2/self` by `clever curl /v2/self`
Your url is always the same, no ? Maybe with on Premise it's not the case. :man_shrugging: 

Can we do something like 
if the url start with `/` then we add the env var `CC_TOOLS_API_ENDPOINT` or `https://api.clever-cloud.com`

chmuche:
@davlgd You Mention 

`clever curl https://api.clever-cloud.com/v4/billing/organisations/<YOUR_ID>/invoices`

but I can't find this endpoint on de openAPI

https://www.clever-cloud.com/doc/openapi
or
https://developers.clever-cloud.com/openapi/

This exactly what I need to automatically integrate the invoice inside our accounting software.

Thanks :heart:

davlgd:
The main concept of `clever curl` command is to allow to use `curl` "as-is" in a Clever Cloud authed context. I'm not sure we want to start to modify the request on the fly. 

About v4/billing endpoints, public documentation step is coming soon ;)

pdesoyres-cc:
Closing this issue as fixed by https://github.com/CleverCloud/clever-tools/pull/632

PR #632, opened 2023-11-22
feat (curl): list command, add help details

### LINKED ISSUE #617: clever curl: print error + help when no params
If a user launches `clever curl` without any url, such an error is printed : 

```
node:internal/validators:162
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
          ^

TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received undefined
    at validateString (node:internal/validators:162:11)
    at Url.parse (node:url:170:3)
    at Object.urlParse [as parse] (node:url:141:13)
    at parseCurlCommand (/usr/local/lib/node_modules/clever-tools/node_modules/curlconverter/util.js:141:25)
    at Object.curl (/usr/local/lib/node_modules/clever-tools/src/commands/curl.js:26:23)
    at Object.<anonymous> (/usr/local/lib/node_modules/clever-tools/bin/clever.js:1096:38)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'ERR_INVALID_ARG_TYPE'
}

Node.js v21.1.0
```

We could use default URL, like `https://api.clever-cloud.com/v2/organisations`

If not, we should print a better error message.

davlgd:
Fix #617 #615

### DIFF
--- a/src/commands/curl.js
+++ b/src/commands/curl.js
@@ -1,9 +1,10 @@
 'use strict';
 
-const { parseCurlCommand } = require('curlconverter/util.js');
 const { spawn } = require('child_process');
 const { loadOAuthConf, conf } = require('../models/configuration.js');
 const { addOauthHeader } = require('@clevercloud/client/cjs/oauth.js');
+const Logger = require('../logger.js');
+const colors = require('colors/safe');
 
 async function loadTokens () {
   const tokens = await loadOAuthConf();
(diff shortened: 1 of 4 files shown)


src/commands/curl.js
```diff
@@ -4,6 +4,8 @@ const { parseCurlCommand } = require('curlconverter/util.js');
 const { spawn } = require('child_process');
 const { loadOAuthConf, conf } = require('../models/configuration.js');
 const { addOauthHeader } = require('@clevercloud/client/cjs/oauth.js');
+const { println } = require('../logger.js');
+const apiUrl = 'https://api.clever-cloud.com';
```
miton18, line 8: can't we resolve it like other commands ?
https://github.com/CleverCloud/clever-tools/blob/3256552d3582f7163405e544cd3eaf57f6036982/src/models/configuration.js#L91

src/commands/curl.js
```diff
@@ -15,16 +17,46 @@ async function loadTokens () {
...
+  const apiDocUrlv2 = 'https://developers.clever-cloud.com/api/v2/';
+  const apiDocUrlv4 = 'https://developers.clever-cloud.com/api/v4/';
+
+  println(`Usage: clever curl
+Query Clever Cloud's API using Clever Tools credentials. For example: 
+
+  clever curl ${apiUrl}/v2/self
+  clever curl ${apiUrl}/v2/summary
+  clever curl ${apiUrl}/v4/products/zones
+  clever curl ${apiUrl}/v2/organisations/<YOUR_ID>/applications | jq '.[].id'
```
miton18, line 30:
```suggestion
  clever curl ${apiUrl}/v2/organisations/<ORGANISATION_ID>/applications | jq '.[].id'
```

src/commands/curl.js
```diff
@@ -15,16 +17,46 @@ async function loadTokens () {
...
+  clever curl ${apiUrl}/v4/billing/organisations/<YOUR_ID>/<INVOICE_NUMBER>.pdf > invoice.pdf
```
miton18, line 31:
```suggestion
  clever curl ${apiUrl}/v4/billing/organisations/<ORGANISATION_ID>/<INVOICE_NUMBER>.pdf > invoice.pdf
```

miton18:
QA OK

miton18 (changes requested):

src/commands/curl.js
```diff
@@ -15,16 +17,46 @@ async function loadTokens () {
   };
 }
 
+function printHelp () {
+  const apiDocUrlv2 = 'https://developers.clever-cloud.com/api/v2/';
+  const apiDocUrlv4 = 'https://developers.clever-cloud.com/api/v4/';
+
+  println(`Usage: clever curl
```
aurrelhebert, line 24: Why not console.log?

aurrelhebert:
@davlgd could you rebase your branch?

aurrelhebert:
Shouldn't we a log on `clever --help` about the `clever curl` command?

src/commands/curl.js
davlgd: I thought we should use the logger for such print, but I can replace it if I misunderstood

davlgd:
> Shouldn't we a log on `clever --help` about the `clever curl` command?

I thought it was impossible, so I tried it (in a separate commit). It works for me, but maybe I'm too optimistic :D

aurrelhebert:
> > Shouldn't we a log on `clever --help` about the `clever curl` command?
> 
> I thought it was impossible, so I tried it (in a separate commit). It works for me, but maybe I'm too optimistic :D

Indeed this does works however, your change on the `clever curl --help` aren't plotted

davlgd:
> > > Shouldn't we a log on `clever --help` about the `clever curl` command?
> > 
> > 
> > I thought it was impossible, so I tried it (in a separate commit). It works for me, but maybe I'm too optimistic :D
> 
> Indeed this does works however, your change on the `clever curl --help` aren't plotted

Good catch, and I can't declare a short/long description for the command. I'm sad. I delete unecessary code in the second commit, when it will be ok for merge, squash them.

davlgd:
BTW, we can't do that: if the user make a request with some parameters, cliparse won't handle them, it's why it was implemented another way before. I revert this.

davlgd:
Branch is rebased, I added JSON format support. For now, in JSON mode the `CC_RUN_COMMAND` won't be added at app creation (as it show a `human` message). But as env set at creation will be available through API soon (I have to check status), I propose to clarify this in documentation for now, and add JSON support in a future PR.

aurrelhebert:
@davlgd LGTM, should we wait for your test? 
QA OK

src/commands/curl.js
```diff
@@ -15,16 +16,46 @@ async function loadTokens () {
...
+
+Our API documentation is available here : 
+
+  ${apiDocUrlv2}
+  ${apiDocUrlv4}`);
+}
+
 async function curl () {
 
+  const curlNeedsHelp = process.argv[2] !== 'curl' || process.argv.length < 3 || process.argv[3] === '--help' || process.argv[3] === '-h';
```
aurrelhebert, line 40: I don't know if it's easy but in that case the --help requires to be set in 3rd position. Shouldn't we plot the help too if customer enters: 

```sh
curl https://api.clever-cloud.com/v2/self --help
```

davlgd: I think the condition is already complex enough, and user can type:

```
clever curl --help https://api.clever-cloud.com/v2/self
```

If we add this condition, it will not work on many case if there is a cURL argument for exemple. And it will be without an end. We control everything before the `curl` we shouldn't try to handle what's coming after

hsablonniere:
@davlgd We changed a lot of details about this (including updating the curl converter). Can you test it again before we merge?

aurrelhebert (approved):

davlgd:
> @davlgd We changed a lot of details about this (including updating the curl converter). Can you test it again before we merge?

LGTM

miton18 (approved):

PR #634, opened 2023-11-22
feat(applications): use current directory as default create name 

### LINKED ISSUE #616: clever create : optional app name
Currently, to create an app, we need to declare its name: 

`clever create -t <app_type> <app_name>`

It could be simpler if this parameter became optional. We could use the current folder name as the default app name and change the "human" message from : 

>Your application has been successfully created!

to 

>Your application has been successfully created as 'app_name'!

davlgd:
Fix #616

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -90,7 +90,10 @@ function run () {
       description: 'Application ID (or name, if unambiguous)',
       parser: Parsers.appIdOrName,
     }),
-    appNameCreation: cliparse.argument('app-name', { description: 'Application name' }),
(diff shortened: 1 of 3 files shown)


hsablonniere:
@davlgd I reworked your commits

* removed the interactivity, we decided it's a bit too much for this usage
* we thought the impl of "default is current directory name" should be in the command, not `clever.js`

This simplifies the code a bit. I kept you as the author :wink:

hsablonniere:
:ballot_box_with_check: I just tested the built preview and it works!

* create without a name
* create with an explicit name

aurrelhebert (approved):

PR #636, opened 2023-11-22
feat: add a --no-color option

### LINKED ISSUE #631: Add a --no-color option
Most CLIs offer a --no-color option to avoid having messages containing color codes. 

Here's an example with a simple Rust system call:
```Rust
stderr: "\u{1b}[1m\u{1b}[31m[ERROR] \u{1b}[39m\u{1b}[22mYou do not belong to this organisation so you cannot access it\n"
```

NOTE: As a temporary fix, you can export FORCE_COLOR=0 to disable the color.

davlgd:
Fix #631

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1,15 +1,34 @@
 #! /usr/bin/env node
 'use strict';
 
+function hasParam (param, paramValue) {
+  const index = process.argv.indexOf(param);
+  if (index === -1) {
+    return false;
+  }
+  if (paramValue != null) {
+    return process.argv[index + 1] === paramValue;
+  }
+  return true;
+}
+
 // These need to be set before Logger and other stuffs
-if (process.argv.includes('-v') || process.argv.includes('--verbose')) {
+if (hasParam('-v') || hasParam('--verbose')) {
   process.env.CLEVER_VERBOSE = '1';
 }
 
 // These need to be set before Logger and other stuffs
 // Don't log anything in autocomplete mode
-if (process.argv.includes('--autocomplete-index')) {
+if (hasParam('--autocomplete-index')) {
   process.env.CLEVER_QUIET = '1';
+}
+
+// These need to be set before other stuffs
+const colors = require('colors');
+const colorExplicitFalse = hasParam('--no-color') || hasParam('--color', 'false');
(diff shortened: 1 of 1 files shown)


bin/clever.js
```diff
@@ -12,6 +12,13 @@ if (process.argv.includes('--autocomplete-index')) {
   process.env.CLEVER_QUIET = '1';
 }
 
+// These need to be set before other stuff
+// Disable color globally if the user asked to
+if (process.argv.includes('--nocolor')) {
```
aurrelhebert, line 17: I think that `no-color` will be better suited. Most of other available option uses a `-` to split words, as for example `--no-update-notifier`.

aurrelhebert (commented):
LGTM, it's working fine for me

bin/clever.js
davlgd: It wasn't working as the parser was seeing 2 different options (if I remember weel). It's why I removed the middle '-'

davlgd: I looked at why I got this behavior, as sometimes the `-` was OK, but not with this option. After several attempts and further analysis, I found that the issue occurs only if an option starts with `--no-`.

For example:

```
$ clever --no-color
Unknown option: color
```

```
$ clever --color-no
Unknown option: color-no
```

For the same reason, the `--no-update-notifier` option can't be used:

```
$ clever logs --no-update-notifier
Unknown option: update-notifier
```

I thought it was a bug/feature of cliparse, but after more deep dive I discovered it's a feature of the `minimist` package used by cliparse. If an argument starts with `--no-` this part is not included in its name and its value is considered as false. You can check that in `minimist` [parsing tests](https://github.com/minimistjs/minimist/blob/main/test/parse.js) and [code](https://github.com/minimistjs/minimist/blob/fa422e82ce6d06278a515343670ebdc161417b90/index.js#L169).

So I pushed an update to deal with this. Now, there is a `--color` option, `true` by default. The user can also use `--no-color` as a shorter form.

davlgd:
ops team asked to implement a quite generic approach for this kind of option: 
- if the app is launched in a non-tty environment: no-color by default
- if the app is launched in a tty environment: color by default
- if the user asks a particular mode, it is given priority

hsablonniere (approved):
LGTM :+1:

PR #648, opened 2023-12-20
fix(addon): skip region check for matomo as not implemented

aurrelhebert:


hsablonniere:
This has been solved by updating the list of add-on providers on the API/data side.

PR #650, opened 2023-12-28
fix (update-notifier): change behavior

### LINKED ISSUE #649: Change --no-update-notifier behavior
As mentioned in #636, the options starting with `--no-` need a special treatment and `--no-update-notifier` is not recognized as a working option. So we should change how the user can ask to not be notified of an update of Clever Tools.

davlgd:
Fix #649 

I checked the code is OK printing the `this.disabled` value of the `update-notifier` package

* [x] Merge https://github.com/CleverCloud/clever-tools/pull/636 first


hsablonniere (approved):
LGTM :+1:

PR #652, opened 2024-01-01
Add .env auto-import (with user confirmation)

### LINKED ISSUE #651: .env import at app creation
When a user creates an app, there is no way to import environment variables automatically. Such a feature would enable easier creation from existing code (open source projects, tutorials, etc.).

As the `.env` file is a common standard, used in many project, we could: 
- detect its presence during ` clever create`
- ask the user if he wants to import its content
- do it if asked

davlgd:
Fix #651 

ATM, there is still an issue to fix. The steps to reproduce are:
- Create an invalid `.env`file
- Try to create with import: an error is raised but not well formatted
- Try to `clever import < .env` the same error is raised but well formatted

There is a same issue if you try to create 2 applications with the same alias: an error is raised but not well formatted.

I think I don't understand something in the error handling, as I encountered the same issue previously working on `clever curl` errors.


aurrelhebert:
Should we really do that? This means that `env` var are stored locally. What would be the vs using a config provider for example?

davlgd:
> Should we really do that? This means that `env` vars are stored locally. What would be the win compared to using a config provider for example?

When a user creates an application or deploy an existing project needing env vars, he should create the app and then import env vars. As we can now do both in 1 step from API, it could be great to have such a feature directly from `clever create` (and it simplifies process for 1-click apps/tutorials). 

Config provider is a Clever Cloud centric way to share env vars with an app, I'm not sure it covers the same needs.

aurrelhebert:
Ok I get the aim, I think that a lot of env var could be shared to an app with a Config Provider enabling already a 1 click order for most tutorials. 

However indeed it could be great to be able to add env variable directly when we order. First I would see something like: 
```sh
clever create --set-env='key=value'
clever create --set-env='key=value'  --set-env='key42=value42' 
```

And from here I am not sure that we should loaded from a file, maybe could something like (we should discuss it I think) : 

```sh
clever create --set-env-from='.env'
clever create --set-env-from='.env.json'
```

aurrelhebert:
Thanks for your feedback on this, the more thing of if, the less I am convinced. I am afraid that this would enforce a sort of bad practice (having a local env file), that could easily lead to leaking secrets. 

I am more in favour to add env vars threw the CLI with `--set-env` and if there is too many threw a configuration provider. Maybe we could see how we can create a Configuration provider from a local file?

PR #657, opened 2024-01-16
fix: add an alias/name check during app create

### LINKED ISSUE #656: prevent link on same alias
Prevent users to link 2 applications with the same alias

this can cause exceptions on other commands

davlgd:
Fix #656

### DIFF
--- a/src/models/app_configuration.js
+++ b/src/models/app_configuration.js
@@ -63,7 +63,7 @@ async function removeLinkedApplication (alias) {
 };
 
 function findApp (config, alias) {
-
+  
   if (_.isEmpty(config.apps)) {
     throw new Error('There are no applications linked. You can add one with `clever link`');
   }
@@ -80,6 +80,20 @@ function findApp (config, alias) {
   }
 
   return findDefaultApp(config);
+}
+
+function checkAlreadyLinked (apps, name, alias) {
+
+  const appAliasExists = apps.some(app => alias && app.alias === alias);
+  const appNameExists = apps.some(app => app.name === name);
+
+  if (appNameExists) {
+    throw new Error(`An application is already linked with the name '${name}'`);
+  }
+
+  if (appAliasExists) {
+    throw new Error(`An application is already linked with the alias '${alias}'`);
+  }
 }
 
 function findDefaultApp (config) {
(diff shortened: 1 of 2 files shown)


src/models/app_configuration.js
```diff
@@ -63,7 +63,7 @@ async function removeLinkedApplication (alias) {
 };
 
 function findApp (config, alias) {
-
+
```
miton18, line 66: useless

src/models/app_configuration.js
```diff
@@ -82,6 +82,20 @@ function findApp (config, alias) {
   return findDefaultApp(config);
 }
 
+function checkAlreadyLinked (apps, name, alias) {
+
+  const appAliasExists = apps.some(app => alias && app.alias === alias);
+  const appNameExists = apps.some(app => app.name === name);
```
miton18, line 88: can you group the expression with the condition ?

miton18 (changes requested):

src/models/app_configuration.js
hsablonniere, line 88: nitpick: we try to alway use parens on arrow functions, can you add them?

```js
  const appAliasExists = apps.some((app) => alias && app.alias === alias);
  const appNameExists = apps.some((app) => app.name === name);
```

src/models/app_configuration.js
```diff
@@ -82,6 +82,20 @@ function findApp (config, alias) {
   return findDefaultApp(config);
 }
 
+function checkAlreadyLinked (apps, name, alias) {
+
+  const appAliasExists = apps.some(app => alias && app.alias === alias);
```
hsablonniere, line 87: nitpick: we try to avoid implicit null test and always use `!= null`. Can add it?

```
  const appAliasExists = apps.some((app) => alias != null && app.alias === alias);
```

hsablonniere (changes requested):
Hey @davlgd, thank you for this. I added a few nitpicks but we tested the branch and it works great!

src/models/app_configuration.js
```diff
@@ -82,6 +81,18 @@ function findApp (config, alias) {
   return findDefaultApp(config);
 }
 
+function checkAlreadyLinked (apps, name, alias) {
+  const appAliasExists = apps.some((app) => app.alias !=null &&  app.alias === alias);
```
miton18, line 85:
```suggestion
  const appAliasExists = apps.some((app) => app.alias != null && app.alias === alias);
```

miton18 (approved):

PR #677, opened 2024-02-09
Add references to new grahds region

davlgd:
- Add `grahds` region in autocomplete
- Add `grahds` region in documentation


miton18:
OK for me, wait for @BlackYoup "prodded" signal

miton18 (approved):

PR #681, opened 2024-02-13
fix(logs): show addon logs when addonId is provided

miton18:


src/commands/logs.js
```diff
@@ -5,24 +5,27 @@ const LogV2 = require('../models/log.js');
 const Log = require('../models/log-v4.js');
 const Logger = require('../logger.js');
 const { Deferred } = require('../models/utils.js');
+const colors = require('colors/safe');
 
 async function appLogs (params) {
-  const { alias, after: since, before: until, search, 'deployment-id': deploymentId } = params.options;
+  const { alias, addon, after: since, before: until, search, 'deployment-id': deploymentId } = params.options;
```
hsablonniere, line 11: suggestion: we could alias it here to `addonId` and reuse it directly on line 22

src/commands/logs.js
```diff
@@ -5,24 +5,27 @@ const LogV2 = require('../models/log.js');
...
   const { appId, ownerId } = await AppConfig.getAppDetails({ alias });
 
-  Logger.println('Waiting for application logs…');
-
   // TODO: drop when addons are migrated to the v4 API
-  if (params.addon) {
+  if (addon) {
+
+  Logger.println(colors.blue('Waiting for addon logs…'));
     const { addon: addonId } = params.options;
```
hsablonniere, line 22: nitpick: no need to get it again from params.options

hsablonniere (approved):
Hey @miton18, thanks for this!!

You changed the color in the same commit, that's OK but I almost missed. I added a suggestion but it's not blocking, feel free to apply it or not and merge + release!

src/commands/logs.js
```diff
@@ -5,24 +5,25 @@ const LogV2 = require('../models/log.js');
...
   const { appId, ownerId } = await AppConfig.getAppDetails({ alias });
 
-  Logger.println('Waiting for application logs…');
-
   // TODO: drop when addons are migrated to the v4 API
-  if (params.addon) {
-    const { addon: addonId } = params.options;
+  if (addonId) {
+    Logger.println(colors.blue('Waiting for addon logs…'));
     const appAddonId = addonId || await AppConfig.getAppDetails({ alias }).then(({ appId }) => appId);
```
hsablonniere, line 21: Do we still need this?

hsablonniere (changes requested):
Last question, sorry.

src/commands/logs.js
miton18: Indeed, dropped

hsablonniere (approved):
LGTM :+1:

Issue #589, opened 2023-07-18
Support `--format` for all commands


hsablonniere:
*  default is human readable (`--format=human`)
  * This is the one with tables, colors, indentation...
* `--format=json` for most commands
* `--format=[FOO]` for specific formats
  * example `clever addon env --format=bash`

aurrelhebert:
For our `JSON` keys we decided to use a lowerCamelCase

pdesoyres-cc:
remaining commands:

- [ ] `activity`
- [ ] `config`
- [ ] `diag`
- [ ] `drain`
- [ ] `env`
- [ ] `notify-email`
- [ ] `profile`
- [ ] `published-configs`
- [ ] `service`
- [ ] `status`
- [ ] `tcp-redirs`
- [ ] `webhooks`

pdesoyres-cc:
some commands support the `--json`/`-j` option. For them, we should make a smooth transition to `--format`/`-F` option. We may support both, but one day, we will remove the `--json`/`-j` option:
- [ ] applications
- [ ] ng/networkgroups

pdesoyres-cc:
About json format and follow behavior.

In some commands, we can follow. for instance in `log` command when no `since` is specified. or in `activity` command.

In `log` command, there is `json-stream` format.
In `activity` command, the json format is not yet supported.

I propose to get rid of `json-stream` format and simplify into:

#### command doesn't imply follow, format is `human`:
```
output line 1
output line 2
```

#### command doesn't imply follow, format is `json`:
```
[
	{"output": "line 1"},
	{"output": "line 2"}
]
```

#### command implies follow, format is `human`:
```
output line 1
output line 2
...
```

#### command implies follow, format is `json`:
```
{"output": "line 1"}
{"output": "line 2"}
...
```

PR #697, opened 2024-02-29
feat(addon): support JSON format when plotting an add-on list

aurrelhebert:
As mentioned in https://github.com/CleverCloud/clever-tools/issues/589 implement format option for the `clever addon` command

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -602,7 +602,7 @@ function run () {
 
   const addonCommands = cliparse.command('addon', {
     description: 'Manage addons',
-    options: [opts.orgaIdOrName],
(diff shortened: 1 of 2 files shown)


miton18:
Current output

```
{
  "addonId": "addon_xxx",
  "creationDate": 1685518606915,
  "name": "test-logs",
  "plan": "XL",
  "realId": "redis_yyy",
  "region": "par",
  "type": "Redis"
}
```

With the human format we output human readable texts like `Redis` for provider type and `XL` for plan.

For the json output, what about using reusable values like `redis-addon` and `xl_mono` (the slug).

miton18 (approved):

PR #701, opened 2024-03-12
feat(backups): support JSON format when plotting a backup list

aurrelhebert:
As mentioned in https://github.com/CleverCloud/clever-tools/issues/589 implement format option for the `clever database backups` command

### DIFF
--- a/src/commands/database.js
+++ b/src/commands/database.js
@@ -8,10 +8,11 @@ const superagent = require('superagent');
 const fs = require('fs');
 const { findOwnerId } = require('../models/addon.js');
 const { resolveRealId } = require('../models/ids-resolver.js');
+const Logger = require('../logger.js');
 
 async function listBackups (params) {
 
-  const { org } = params.options;
+  const { org, format } = params.options;
   const [addonIdOrRealId] = params.args;
 
   const addonId = await resolveRealId(addonIdOrRealId);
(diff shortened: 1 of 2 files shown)


miton18:
QA OK
`jq` accept the output

src/commands/database.js
```diff
@@ -8,40 +8,58 @@ const superagent = require('superagent');
...
-    'STATUS',
-  ];
-
-  println(formatTable([
-    head,
-    ...formattedLines,
-  ]));
+  switch (format) {
+    case 'json': {
+      const formattedBackups = backups.map((backup) => {
```
miton18, line 30: Maybe we can sort on both output ?

src/commands/database.js
```diff
@@ -8,40 +8,58 @@ const superagent = require('superagent');
...
+        return {
+          addon_id: addonId,
```
miton18, line 32: In JS-on world, it's more standard to see object keys in a lowerCamelCase format

src/commands/database.js
aurrelhebert: I am not a huge of sorting both outputs, as the current way to print those lists are in prod since a few time now. An if as me you build a parser on top of the current CLI output we will introduce a breaking change

src/commands/database.js
aurrelhebert: Indeed I didn't saw any `_` in our current JSON output. I am changing it

miton18 (approved):

PR #705, opened 2024-03-27
feat(database-backups): rework json backup list

aurrelhebert:


miton18:
QA OK
works with:
* mysql
* postgresql 

addonID or ID

PR #720, opened 2024-04-12
fix: add-on env export in shell format

### LINKED ISSUE #719: clever addon env -F shell not working in 3.5.x
In release 3.5 we introduced `--format` option for add-on list, but we had such an option for `addon env`. Thus, there is a conflit between these two commands and we can't use `addon env -F shell` anymore

davlgd:
Fix #719

My solution is using the `--add-export` we already use in `clever env` for applications. Thus, it's consistent with current CLI behavior.


hsablonniere:
I'm going to go in favor of https://github.com/CleverCloud/clever-tools/pull/721. This pushes the usage of `-F` in all commands. We'll figure out a good solution for cliparse later.

PR #721, opened 2024-04-13
fix(addon): allow json format in new list subcommand

### LINKED ISSUE #719: clever addon env -F shell not working in 3.5.x

aurrelhebert:
closes #719

This PR revert the JSON format introduced for the `clever addon` command in https://github.com/CleverCloud/clever-tools/pull/697. This introduces a new `clever addon list` command. This sub-command could use a custom format (human or json) to plot all add-ons of an owner. 

To handle option already set in main command in a sub-command, we would need to work on a cliparse patch to allow an option overwrites.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -80,11 +80,11 @@ function run () {
   // ARGUMENTS
   const args = {
     addonIdOrName: cliparse.argument('addon-id', {
-      description: 'Addon ID (or name, if unambiguous)',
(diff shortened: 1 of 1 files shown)


hsablonniere (approved):
LGTM :+1:

PR #726, opened 2024-04-17
feat(addon create): choose the cheapest plan by default

### LINKED ISSUE #653: clever addon create: use cheapest/lowest plan by default
## Context

Right now `clever addon create` uses `dev` plan by default when not specified, it only works for PG and MySQL. We should use the cheapest/lowest plan by default.

pdesoyres-cc:
Fixes #653

### DIFF
--- a/src/models/addon.js
+++ b/src/models/addon.js
@@ -26,7 +26,7 @@ async function getProvider (providerName) {
   const providers = await listProviders();
   const provider = providers.find((p) => p.id === providerName);
   if (provider == null) {
-    throw new Error('invalid provider name');
+    throw new Error(`Invalid provider name. Available providers: ${providers.map((p) => p.id).join(', ')}`);
   }
   return provider;
 }
--- a/src/models/addon.js
+++ b/src/models/addon.js
@@ -136,21 +136,16 @@ function validateAddonVersionAndOptions (region, version, addonOptions, provider
 async function create ({ ownerId, name, providerName, planName, region, skipConfirmation, version, addonOptions }) {
 
   // TODO: We should be able to use it without {}
-  const providers = await listProviders();
+  const provider = await getProvider(providerName);
 
-  const provider = providers.find((p) => p.id === providerName);
-  if (provider == null) {
-    throw new Error('invalid provider name');
-  }
   if (!provider.regions.includes(region)) {
-    throw new Error(`invalid region name. Available regions: ${provider.regions.join(', ')}`);
+    throw new Error(`Invalid region name. Available regions: ${provider.regions.join(', ')}`);
+  }
+  if (provider.plans.length === 0) {
+    throw new Error(`No plans available for provider ${providerName}`);
   }
 
-  const plan = provider.plans.find((p) => p.slug.toLowerCase() === planName.toLowerCase());
-  if (plan == null) {
-    const availablePlans = provider.plans.map((p) => p.slug);
-    throw new Error(`invalid plan name. Available plans: ${availablePlans.join(', ')}`);
-  }
+  const plan = getPlan(planName, provider.plans);
 
   const providerInfos = await getProviderInfos(provider.id);
   const planType = plan.features.find(({ name }) => name.toLowerCase() === 'type');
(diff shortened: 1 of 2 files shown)


aurrelhebert (approved):

aurrelhebert:
QA ok :heavy_check_mark: it's working like a charm

src/models/addon.js
```diff
@@ -289,6 +285,24 @@ function parseAddonOptions (options) {
   }, {});
 }
 
+async function getPlan (planName, provider) {
```
hsablonniere, line 288: nitpick: this function does not need to be async

hsablonniere: nitpick: we could pass the list of plans instead of the whole provider

src/models/addon.js
```diff
@@ -289,6 +285,24 @@ function parseAddonOptions (options) {
...
+  // if no plan specified, pick the cheapest one
+  if (planName == null || planName === '') {
+    if (provider.plans.length === 0) {
+      throw new Error('no plans available');
```
hsablonniere, line 292: question: this will be displayed to the user, is it descriptive enough?

src/models/addon.js
```diff
@@ -289,6 +285,24 @@ function parseAddonOptions (options) {
...
+    }
+
+    return provider.plans.sort((p1, p2) => p1.price - p2.price)[0];
+  }
+
+  const plan = provider.plans.find((p) => p.slug.toLowerCase() === planName.toLowerCase());
+  if (plan == null) {
+    const availablePlans = provider.plans.map((p) => p.slug);
+    throw new Error(`invalid plan name. Available plans: ${availablePlans.join(', ')}`);
```
hsablonniere, line 301: nitpick: we don't have copy guidelines but I would have used an upper case letter at the beginning and a coma before the "available plans:".

hsablonniere (changes requested):
Well done @pdesoyres-cc :clap: 

Tested with the built version on:

* PG no plan
* PG with invalid plan
* PG with valid plan
* redis no plan
* kv no plan
* keycloak no plan
* cellar no plan
* config provider

I added a few comments here and there. I wonder if we should display (human and JSON) the selected plan since we may have selected it for the user.

src/models/addon.js
pdesoyres-cc: This was not the subject of this PR, but I can do that.

src/models/addon.js
pdesoyres-cc: This was not the subject of this PR, but I can do that.

pdesoyres-cc: Do you have a more descriptive suggestion?

pdesoyres-cc: I have introduced this if branch for code robustness. But, this should never happen if the API always returns plans for a available provider (which, as for now, is the case for every providers).

hsablonniere: Let's discuss it together in a meeting and move on :wink:

src/models/addon.js
hsablonniere: Ah you're right, sorry. We'll discuss it in the meeting :+1:

src/models/addon.js
```diff
@@ -136,17 +136,16 @@ function validateAddonVersionAndOptions (region, version, addonOptions, provider
...
   if (!provider.regions.includes(region)) {
-    throw new Error(`invalid region name. Available regions: ${provider.regions.join(', ')}`);
+    throw new Error(`Invalid region name. Available regions: ${provider.regions.join(', ')}`);
+  }
+  if (provider.plans.length === 0) {
+    throw new Error(`No plans available for provider ${providerName}`);
   }
 
-  const plan = await getPlan(planName, provider);
+  const plan = await getPlan(planName, provider.plans);
```
hsablonniere, line 148: issue: you can remove the `await`

hsablonniere (changes requested):

src/models/addon.js
```diff
@@ -289,6 +284,20 @@ function parseAddonOptions (options) {
   }, {});
 }
 
+function getPlan (planName, plans) {
```
miton18, line 287:
```suggestion
function getDefaultPlan (planName, plans) {
```

miton18 (approved):

hsablonniere (approved):

PR #733, opened 2024-04-23
chore: make bad commands exit with 1

### LINKED ISSUE #590: Make bad commands exit with 1
Right now they exit with 0. We will need to update cliparse for this.

pdesoyres-cc:
Fixes #590


package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@clevercloud/client": "^8.1.2",
     "clf-date": "^0.2.0",
-    "cliparse": "^0.3.3",
+    "cliparse": "file:cliparse-0.3.4.tgz",
```
aurrelhebert, line 31: Shouldn't we made a proper update of cliparse?

pdesoyres-cc: Yes, of course, This PR is in a state that lets you check the preview. Once validated, I will make the `cliparse` release and include it properly in the package.json.

aurrelhebert: For the QA which commands are impacted by this change? All?

pdesoyres-cc: This is about bad commands: 
- unknown command should exit with 1
- non-terminated command should exit 1,
- command with missing required arg/option should exit 1,
- command with unknown arg/option should exit 1,
- command with malformed arg/option should exit 1.

aurrelhebert: Ok I see, for instance: 

```
clever wrong_command
./clever addon list --format wrong
```

Shouldn't we then plot an `Error` message? as `[ERROR] format: The output format must be one of human, json`?

aurrelhebert:
I have a run a couple of tests and LGTM, we do get an exit code of 1 any time an error occurs, and 0 otherwise.

hsablonniere (commented):
Well done @pdesoyres-cc :clap: :tada:

I tested the built version in various scenarios and it works (as in triggers and error :stuck_out_tongue:) perfectly

package.json
aurrelhebert: https://github.com/CleverCloud/clever-tools/issues/742

aurrelhebert (approved):

PR #738, opened 2024-05-16
Nix support: new repository

ptitfred:
The original Nix bindings were supported by a company that no longer exists.

I've updated to 3.6.1 at this occasion.

### DIFF
--- a/docs/setup-systems.md
+++ b/docs/setup-systems.md
@@ -155,4 +155,4 @@ COPY --from=clevercloud/clever-tools /bin/clever /usr/local/bin/clever
 
 If you are using Nix, you will find a Nix derivation on Fretlink's GitHub repository:
 
-* https://github.com/fretlink/clever-tools-nix
+* https://github.com/funky-thunks/clever-tools-nix


hsablonniere:
Hello @ptitfred, thanks for you PR, as I don't really know much about Nix, I sent a ping to my Nix powered colleagues :wink: and we'll get back to you.

florian-sanders-cc:
Hello @ptitfred

We've actually been considering how to distribute the CLI package with Nix so your PR timing is definitely on point!
Thanks for this :wink:

We would prefer to move away from third party repositories and contribute the CLI package directly to the nixpkgs repository.

This would make it easier for users to install it and we would be in charge of maintaining the package.

We would also like to change the package code because Nix has evolved since the first versions of the CLI and there are new ways to package Node.js projects.

The package would rely on the Nix `buildNpmPackage` helper instead of patching the prebuilt Linux binary. There are several advantages to this:
- The package code is very easy to maintain, in fact it could even be maintained without much Nix knowledge,
- The package should work for every system supported by Nix instead of Linux only (since some people use Nix on MacOS for instance, it's a nice bonus to support this).

We'll close this PR and track this subject within a dedicated issue (see #748).
This way you can subscribe to it if you want to get notified when the package is available on Nixpkgs.

Thanks again for your help!

ptitfred:
Ok.

Is it merged in nixpkgs already?

Currently the link in the readme leads to code that doesn't work anymore. If it's merged in nixpkgs a PR should link to it then. If it's not merged in nixpkgs, I'd suggest that instead we drop the Nix mention in the readme.

Another option in the meantime would have been to merge this PR even if a better option is coming soon.

florian-sanders-cc:
The PR within nixpkgs should be done this week so it's not merged indeed.

We discussed it with the CLI team and we'll drop this section from the docs for the moment.

Thanks for pointing it out.

ptitfred:
> We discussed it with the CLI team and we'll drop this section from the docs for the moment.

This section was outdated for so long it won't be missed I suppose.

Anyway good to hear it would be upstreamed in nixpkgs in the end :tada: always the best option

hsablonniere:
Fixed in #787 !

ptitfred:
> Fixed in #787 !

Thank you

PR #740, opened 2024-05-23
fix(drain): wrong link between command option and its processing

KannarFr:


hsablonniere (approved):
I haven't tested it but code seems OK :+1:

juliamrch:
I've tested it, works like a charm.

PR #744, opened 2024-05-23
feat: list all applications

### LINKED ISSUE #737: List an owner applications
We should be able to list an owner applications. 

This is currently possible threw the following `cURL` command:

```
clever curl https://api.clever-cloud.com/v2/organisations/<owner_id>/applications | jq '.[].name'
```

davlgd:
fixes #737

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -660,9 +660,14 @@ function run () {
 
   // APPLICATIONS COMMAND
   const applications = lazyRequirePromiseModule('../src/commands/applications.js');
+  const applicationsListRemoteCommand = cliparse.command('list', {
+    description: 'List all applications',
+    options: [opts.orgaIdOrName, opts.humanJsonOutputFormat],
+  }, applications('listAll'));
   const applicationsCommand = cliparse.command('applications', {
     description: 'List linked applications',
-    options: [opts.onlyAliases, opts.jsonFormat],
+    privateOptions: [opts.onlyAliases, opts.jsonFormat],
+    commands: [applicationsListRemoteCommand],
   }, applications('list'));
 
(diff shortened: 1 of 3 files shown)


aurrelhebert:
We could discuss about something to have the same behavior for: 
```sh
clever applications list
clever addon list
```

On usage, I would prefer a `clever applications list --org orga_XXX` over `clever applications --all orga_XXX`

For the rest: 
  - I really like the Human view, really nice job!
  - LGTM for the JSON format

davlgd:
@hsablonniere I'll let you check everything is ok with your test org. It seems fine here.

aurrelhebert:
LGTM, only one tiny question, should we update the --org description for this command (as it include an alternative behavior?)

davlgd:
> LGTM, only one tiny question, should we update the --org description for this command (as it includes an alternative behavior?)

```
Organisation to target by its ID (or name, if unambiguous)
```

Seems good to me. If we want to change it we need to create another option, maybe we'll be able to be more precise in such situations with the new options mechanism.

hsablonniere (approved):
Nice!!! LGTM :+1:

hsablonniere (changes requested):
ABORT! There's a bug.

When you run `clever applications` in a repo where you have linked apps but not with the current account, it's empty.

hsablonniere (approved):
@aurrelhebert @davlgd @miton18 @pdesoyres-cc

We realized it was not possible to mix the existing list with the new one without messing up the old/new JSON format and having weird edge cases with the linked apps.

This PR now proposes a new `list` subcommand on `clever applications` with only `-F json` and `-o NAME/ID` to filter. The old `clever applications` command is kept as is.

PR #749, opened 2024-05-28
feat(parsers): add a parser to parse boolean with a default value true

aurrelhebert:


src/parsers.js
```diff
@@ -191,6 +191,20 @@ function durationInSeconds (durationStr = '') {
   }
 }
 
+/**
+ * Parse a boolean input which is expect a default value as true
+ * An empty value returns true
+ * @returns {bool} true or false
```
pdesoyres-cc, line 197: `{bool}` type doesn't exist. you should use `{boolean}` instead.

pdesoyres-cc (changes requested):
LGTM with just a little comment

miton18:
Do we need a bool parser ?

aurrelhebert:
Closing as related to the `watch` param we decided not to support

PR #753, opened 2024-05-30
chore: use strict version specifiers in package.json

hsablonniere:


pdesoyres-cc:
Can you say why this is needed?

aurrelhebert (approved):

Issue #748, opened 2024-05-27
Nix packaging: move to nixpkgs


florian-sanders-cc:
## Context

The [project documentation](https://github.com/CleverCloud/clever-tools/blob/b26819f4ddd21ce43908fbede0321d03eef9880f/docs/setup-systems.md#nix-package-manager) links to a third party repository to install the CLI with Nix.

As pointed out in #738, this repository is not maintained and would need to be updated.

We have been considering the subject of nix packaging lately and we think it would be better to move away from third party repositories and contribute the CLI package directly to the [nixpkgs repository](https://github.com/NixOS/nixpkgs).

This would make it easier for users to install it and we would be in charge of maintaining the package.

We would also like to change the package code because Nix has evolved since the first versions of the CLI and there are new ways to package Node.js projects.

The package would rely on the Nix `buildNpmPackage` helper instead of patching the prebuilt Linux binary.
There are several advantages to this:
- The package code is very easy to maintain, in fact it could even be maintained without much Nix knowledge,
- The package should work for every system supported by Nix instead of Linux only (since some people use Nix on MacOS for instance, it's a nice bonus to support this).

An example of the code is available as a flake on a [separate personal repo](https://github.com/florian-sanders-cc/clever-tools-flake/blob/master/flake.nix#L25) ( :warning:  this is mostly a POC, do not use it as it will be deleted soon).

It should be easy to transform it for nixpkgs.

florian-sanders-cc:
The PR is ready to be submitted on Nixpkgs but we'll wait for the next Clever Tools release (coming very soon) because it fixes the build for MacOS (see #754).

florian-sanders-cc:
PR has been created : https://github.com/NixOS/nixpkgs/pull/319818

florian-sanders-cc:
PR has been merged :tada: 
I've also closed an (old) issue from a customer on nixpkgs: https://github.com/NixOS/nixpkgs/issues/163506

I'll create a PR on this repo to update the docs about nix installation :+1:

PR #754, opened 2024-05-31
fix: fix clever curl in packaged and node version

hsablonniere:


davlgd:
This version is OK for macOS (Apple Silicon)

florian-sanders-cc:
Not that it matters but this also fixes the upcoming Nix package on MacOS Intel so thanks for this :heart_eyes: 

I'll wait for this to be released before moving forward with the Nix package (#748).

PR #755, opened 2024-06-03
feat: support --format for all commands

### LINKED ISSUE #589: Support `--format` for all commands

pdesoyres-cc:
Fixes #589

* [x] release and update cliparse before https://github.com/CleverCloud/cliparse-node/pull/35

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1057,6 +1057,7 @@ function run () {
   const tcpRedirsCommands = cliparse.command('tcp-redirs', {
     description: 'Control the TCP redirections from reverse proxies to your application',
     options: [opts.alias, opts.appIdOrName],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [tcpRedirsListNamespacesCommand, tcpRedirsAddCommand, tcpRedirsRemoveCommand],
   }, tcpRedirs('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -776,6 +776,7 @@ function run () {
   const drainCommands = cliparse.command('drain', {
     description: 'Manage drains',
     options: [opts.alias, opts.appIdOrName, opts.addonId],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [drainCreateCommand, drainRemoveCommand, drainEnableCommand, drainDisableCommand],
   }, drain('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -956,6 +956,7 @@ function run () {
   const profile = lazyRequirePromiseModule('../src/commands/profile.js');
   const profileCommand = cliparse.command('profile', {
     description: 'Display the profile of the current user',
+    options: [opts.humanJsonOutputFormat],
   }, profile('profile'));
 
   // PUBLISHED CONFIG COMMANDS
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1090,6 +1090,7 @@ function run () {
   const webhooksCommand = cliparse.command('webhooks', {
     description: 'Manage webhooks',
     options: [opts.orgaIdOrName, opts.listAllNotifications],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [addWebhookCommand, removeWebhookCommand],
   }, webhooks('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -724,6 +724,7 @@ function run () {
   const diagCommand = cliparse.command('diag', {
     description: 'Diagnose the current installation (prints various informations for support)',
     args: [],
+    options: [opts.humanJsonOutputFormat],
   }, diag('diag'));
 
   // DOMAIN COMMANDS
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1015,6 +1015,7 @@ function run () {
   const serviceCommands = cliparse.command('service', {
     description: 'Manage service dependencies',
     options: [opts.alias, opts.appIdOrName, opts.onlyApps, opts.onlyAddons, opts.showAll],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [serviceLinkAppCommand, serviceUnlinkAppCommand, serviceLinkAddonCommand, serviceUnlinkAddonCommand],
   }, service('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -976,6 +976,7 @@ function run () {
   const publishedConfigCommands = cliparse.command('published-config', {
     description: 'Manage the configuration made available to other applications by this application',
     options: [opts.alias, opts.appIdOrName],
+    privateOptions: [opts.envFormat],
     commands: [publishedConfigSetCommand, publishedConfigRemoveCommand, publishedConfigImportCommand],
   }, publishedConfig('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1044,6 +1044,7 @@ function run () {
   const tcpRedirs = lazyRequirePromiseModule('../src/commands/tcp-redirs.js');
   const tcpRedirsListNamespacesCommand = cliparse.command('list-namespaces', {
     description: 'List the namespaces in which you can create new TCP redirections',
+    options: [opts.humanJsonOutputFormat],
   }, tcpRedirs('listNamespaces'));
   const tcpRedirsAddCommand = cliparse.command('add', {
     description: 'Add a new TCP redirection to the application',
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -934,6 +934,7 @@ function run () {
   const emailNotificationsCommand = cliparse.command('notify-email', {
     description: 'Manage email notifications',
     options: [opts.orgaIdOrName, opts.listAllNotifications],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [addEmailNotificationCommand, removeEmailNotificationCommand],
   }, notifyEmail('list'));
 
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -638,6 +638,7 @@ function run () {
   const addonProvidersCommand = cliparse.command('providers', {
     description: 'List available add-on providers',
     commands: [addonShowProviderCommand],
+    options: [opts.humanJsonOutputFormat],
   }, addon('listProviders'));
   const addonEnvCommand = cliparse.command('env', {
     description: 'List environment variables for an add-on',
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -155,6 +155,7 @@ function run () {
   const opts = {
     sourceableEnvVarsList: cliparse.flag('add-export', { description: 'Display sourceable env variables setting' }),
     logsFormat: getOutputFormatOption(['json-stream']),
+    activityFormat: getOutputFormatOption(['json-stream']),
     envFormat: getOutputFormatOption(['shell']),
     accesslogsFollow: cliparse.flag('follow', {
       aliases: ['f'],
@@ -612,7 +613,7 @@ function run () {
   const activity = lazyRequirePromiseModule('../src/commands/activity.js');
   const activityCommand = cliparse.command('activity', {
     description: 'Show last deployments of an application',
-    options: [opts.alias, opts.appIdOrName, opts.follow, opts.showAllActivity],
+    options: [opts.alias, opts.appIdOrName, opts.follow, opts.showAllActivity, opts.activityFormat],
   }, activity('activity'));
 
   // ADDON COMMANDS
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -155,7 +155,7 @@ function run () {
   const opts = {
     sourceableEnvVarsList: cliparse.flag('add-export', { description: 'Display sourceable env variables setting' }),
     logsFormat: getOutputFormatOption(['json-stream']),
-    addonEnvFormat: getOutputFormatOption(['shell']),
+    envFormat: getOutputFormatOption(['shell']),
     accesslogsFollow: cliparse.flag('follow', {
       aliases: ['f'],
       description: 'Display access logs continuously (ignores before/until, after/since)',
@@ -642,7 +642,7 @@ function run () {
   }, addon('listProviders'));
   const addonEnvCommand = cliparse.command('env', {
     description: 'List environment variables for an add-on',
-    options: [opts.addonEnvFormat],
+    options: [opts.envFormat],
     args: [opts.addonId],
   }, addon('env'));
   const addonListCommand = cliparse.command('list', {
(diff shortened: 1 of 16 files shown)


aurrelhebert (commented):
LGTM, thanks for your PR! This is a really nice job. I reviewed your code however I didn't run a QA yet

bin/clever.js
```diff
@@ -763,6 +768,7 @@ function run () {
   const envCommands = cliparse.command('env', {
     description: 'Manage environment variables of an application',
     options: [opts.alias, opts.sourceableEnvVarsList],
+    privateOptions: [opts.envFormat],
```
davlgd, line 771: same as above

bin/clever.js
```diff
@@ -895,6 +901,7 @@ function run () {
   const emailNotificationsCommand = cliparse.command('notify-email', {
     description: 'Manage email notifications',
     options: [opts.orgaIdOrName, opts.listAllNotifications],
+    privateOptions: [opts.humanJsonOutputFormat],
```
davlgd, line 904: same as above

bin/clever.js
```diff
@@ -935,6 +943,7 @@ function run () {
   const publishedConfigCommands = cliparse.command('published-config', {
     description: 'Manage the configuration made available to other applications by this application',
     options: [opts.alias],
+    privateOptions: [opts.envFormat],
```
davlgd, line 946: same as above

bin/clever.js
```diff
@@ -973,6 +982,7 @@ function run () {
   const serviceCommands = cliparse.command('service', {
     description: 'Manage service dependencies',
     options: [opts.alias, opts.onlyApps, opts.onlyAddons, opts.showAll],
+    privateOptions: [opts.humanJsonOutputFormat],
```
davlgd, line 985: same as above

bin/clever.js
```diff
@@ -1045,6 +1057,7 @@ function run () {
   const webhooksCommand = cliparse.command('webhooks', {
     description: 'Manage webhooks',
     options: [opts.orgaIdOrName, opts.listAllNotifications],
+    privateOptions: [opts.humanJsonOutputFormat],
```
davlgd, line 1060: same as above

bin/clever.js
```diff
@@ -1014,6 +1025,7 @@ function run () {
   const tcpRedirsCommands = cliparse.command('tcp-redirs', {
     description: 'Control the TCP redirections from reverse proxies to your application',
     options: [opts.alias],
+    privateOptions: [opts.humanJsonOutputFormat],
```
davlgd, line 1028: same as above

src/commands/addon.js
```diff
@@ -156,18 +156,42 @@ async function rename (params) {
...
-  Logger.println(formatTable(formattedProviders));
+  switch (format) {
+    case 'json': {
+      const formattedProviders = providers.map((provider) => ({
+        id: provider.id,
+        name: provider.name,
+        shortDesc: provider.shortDesc,
+        logoUrl: provider.logoUrl,
+        status: provider.status,
+        canUpgrade: provider.canUpgrade,
```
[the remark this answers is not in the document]
davlgd: +1

src/commands/addon.js
```diff
@@ -156,18 +156,42 @@ async function rename (params) {
...
-    ];
-  });
-  Logger.println(formatTable(formattedProviders));
+  switch (format) {
+    case 'json': {
+      const formattedProviders = providers.map((provider) => ({
+        id: provider.id,
+        name: provider.name,
+        shortDesc: provider.shortDesc,
+        logoUrl: provider.logoUrl,
```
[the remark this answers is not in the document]
davlgd: I agree, we can remove it. Status can be removed too (it's always `release` even for Materia KV for example).

src/commands/domain.js
```diff
@@ -26,17 +26,29 @@ function getFavouriteDomain ({ ownerId, appId }) {
...
+
+  switch (format) {
+    case 'json': {
+      Logger.println(formattedDomains);
+      break;
+    }
+    case 'human':
+    default: {
+      formattedDomains.forEach(({ fqdn, favourite }) => {
+        Logger.println(`${favourite ? '*' : ' '} ${fqdn}`);
```
davlgd, line 48: By default, this leads to ' domain.cleverapps.io' with an empty space at the beginning. I suggest to replace it with:

```suggestion
        Logger.println(`${fqdn}${favourite ? color.green(' (favourite)') : ''} `);
```

(it will need `const color = require('colors/safe');`)

src/commands/profile.js
```diff
@@ -5,15 +5,40 @@ const colors = require('colors/safe');
...
+    lang: user.lang,
+    emailValidated: user.emailValidated,
+    has2FA: (user.preferredMFA != null && user.preferredMFA !== 'NONE'),
+  };
+
+  switch (format) {
+    case 'json': {
+      Logger.printJson(formattedUser);
+      break;
+    }
```
davlgd, line 32: Maybe we could just reply with full user data ?

```suggestion
  switch (format) {
    case 'json': {
      Logger.printJson(user);
      break;
    }
```

src/commands/profile.js
```diff
@@ -5,15 +5,40 @@ const colors = require('colors/safe');
...
+    case 'human':
+    default: {
+      Logger.println('You\'re currently logged in as:');
+      Logger.println(`User id          ${formattedUser.id}`);
+      Logger.println(`Name             ${formattedUser.name ?? colors.red.bold('[not specified]')}`);
+      Logger.println(`Email            ${formattedUser.email}`);
+      Logger.println(`Two factor auth  ${formattedUser.has2FA ? 'yes' : 'no'}`);
+    }
```
davlgd, line 40: Here I would add some infos (and adapt to change above):

```suggestion
      Logger.println(`User id          ${user.id}`);
      Logger.println(`Name             ${user.name ?? colors.red.bold('[not specified]')}`);
      Logger.println(`Email            ${user.email}`);
      Logger.println(`OAuth apps       ${user.oauthApps.join(', ')}`);
      Logger.println(`Two factor auth  ${user.has2FA ? 'yes' : 'no'}`);
      Logger.println(`Admin            ${user.admin ? 'yes' : 'no'}`);
      Logger.println(`Can pay?         ${user.canPay ? 'yes' : 'no'}`);
    }
```

src/commands/tcp-redirs.js
```diff
@@ -10,24 +10,45 @@ const Logger = require('../logger.js');
...
+    case 'json': {
+      Logger.printJson(formattedNamespaces);
+      break;
+    }
+    case 'human':
+    default: {
+      Logger.println('Available namespaces: ' + formattedNamespaces.join(', '));
+    }
+  }
+}
```
davlgd, line 28: As namespaces are not always understood well, I suggest to add some information in `human` format for `default`/`cleverapps`

```suggestion
  namespaces.sort((a, b) => a.namespace.localeCompare(b.namespace));

  switch (format) {
    case 'json': {
      Logger.println(JSON.stringify(namespaces, null, 2));
      break;
    }
    case 'human':
    default: {
      Logger.println('Available namespaces:');
      namespaces.forEach(({ namespace }) => {
        switch (namespace) {
          case 'cleverapps':
            Logger.println(`- ${namespace}: for redirections used with 'cleverapps.io' domain`);
            break;
          case 'default':
            Logger.println(`- ${namespace}: for redirections used with custom domains`);
            break;
          default:
            Logger.println(`- ${namespace}`);
        }
      });
      break;
    }
  }
}
```

davlgd (changes requested):
Great job! It seems there are problems with `format` options when it's private on commands used by default. 

I've added some suggestions/nitpicks

src/commands/env.js
```diff
@@ -17,19 +18,47 @@ async function list (params) {
...
+        })),
+        fromDependencies: envFromDeps.map((dep) => ({
+          addonId: dep.app_id,
+          addonName: dep.app_name,
+          env: dep.env,
+        })),
+      });
+      break;
+    }
+    case 'shell':
```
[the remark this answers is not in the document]
pdesoyres-cc: shell and human format use the same algorithm. The only difference is at line 45

bin/clever.js
```diff
@@ -618,6 +619,7 @@ function run () {
   const addonCommands = cliparse.command('addon', {
     description: 'Manage add-ons',
     options: [opts.orgaIdOrName],
+    privateOptions: [opts.humanJsonOutputFormat],
     commands: [addonCreateCommand, addonDeleteCommand, addonRenameCommand, addonListCommand, addonProvidersCommand, addonEnvCommand],
   }, addon('list'));
```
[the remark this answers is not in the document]
pdesoyres-cc: `--format` option is private here so it doesn't apply to any of the child commands. That's why the `list` sub command also has the `--format` option.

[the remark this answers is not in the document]
pdesoyres-cc: removing the default command is a breaking change.

src/commands/activity.js
```diff
@@ -67,27 +78,61 @@ function clearPreviousLine () {
...
+    case 'human':
+    default: {
+      return {
+        print: (event) => Logger.println(formatActivityLine(event)),
+        pop: clearPreviousLine,
+        end: () => {},
+      };
+    }
+  }
 }
```
hsablonniere, line 132: note: not that easy to read/understand but it's mainly because of how the follow mode works with the erase line etc... Let's go with it but I think we'll need to reconsider the behaviour follow mode in the future.

bin/clever.js
```diff
@@ -578,7 +578,7 @@ function run () {
   const activity = lazyRequirePromiseModule('../src/commands/activity.js');
   const activityCommand = cliparse.command('activity', {
     description: 'Show last deployments of an application',
-    options: [opts.alias, opts.follow, opts.showAllActivity],
+    options: [opts.alias, opts.follow, opts.showAllActivity, opts.humanJsonOutputFormat],
```
hsablonniere, line 581: issue: I know you don't agree with @miton18 (and maybe me) on this but in logs and access-logs commands, `json` format produces valid JSON and not NDJSON. I know `jq` will handle both but I think we should discuss the naming.

Also, unlike logs and access-logs that can stream and end, `clever activity -f` will virtually never end.

hsablonniere (changes requested):
Impressive work @pdesoyres-cc :clap: Well done !!! :+1: 

I added some comments here and there. We need to discuss some details (format and naming) but overall this is awesome!

I think the cliparse private options works as expected. Maybe @davlgd didn't have the latest version or local cliparse.

I think we should remove the format support for the domain list command for now as @florian-sanders-cc will soon introduce new domain commands and it will be a good opportunity to return properties like `host` (real host without path prefix), `pathPrefix` and maybe a property for the combination of both.

src/commands/profile.js
davlgd: So maybe we should add `avatar`, `oAuthApps`, `admin` and send same infos in both formats as mentioned below

src/commands/addon.js
```diff
@@ -196,48 +196,81 @@ async function listProviders (params) {
...
-            Logger.println(`    ${name}: default=${enabled}`);
+
+  const formattedProvider = {
+    id: provider.id,
+    name: provider.name,
+    shortDesc: provider.shortDesc,
+    logoUrl: provider.logoUrl,
+    status: provider.status,
+    canUpgrade: provider.canUpgrade,
+    regions: provider.regions,
```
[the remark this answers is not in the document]
pdesoyres-cc: I didn't know that. That's why I didn't change the initial behavior. Maybe we should create a separate issue for that?

src/commands/domain.js
```diff
@@ -26,17 +26,29 @@ function getFavouriteDomain ({ ownerId, appId }) {
...
   const favouriteDomain = await getFavouriteDomain({ ownerId, appId });
-  return app.vhosts.forEach(({ fqdn }) => {
-    const prefix = (fqdn === favouriteDomain)
-      ? '* '
-      : '  ';
-    Logger.println(prefix + fqdn);
-  });
+
+  const formattedDomains = app.vhosts.map(({ fqdn }) => ({
+    fqdn,
```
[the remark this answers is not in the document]
pdesoyres-cc: Maybe @davlgd will have an idea.

src/commands/drain.js
```diff
@@ -15,28 +15,43 @@ async function getAppOrAddonId ({ alias, addonId }) {
...
+      drains.forEach((drain) => {
+        const { id, state, target } = drain;
+        const { url, drainType, indexPrefix, structuredDataParameters } = target;
+
+        let drainView = `${id} -> ${state} for ${url} as ${drainType}`;
+        if (indexPrefix != null) {
+          drainView += `, index: '${indexPrefix}-<YYYY-MM-DD>'`;
+        }
+        if (structuredDataParameters != null) {
+          drainView += `, sd-parameters: '${structuredDataParameters}'`;
```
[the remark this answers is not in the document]
pdesoyres-cc: thanks

bin/clever.js
pdesoyres-cc: I am very open to discussion on that point.

src/commands/addon.js
```diff
@@ -156,18 +156,42 @@ async function rename (params) {
...
+  switch (format) {
+    case 'json': {
+      const formattedProviders = providers.map((provider) => ({
+        id: provider.id,
+        name: provider.name,
+        shortDesc: provider.shortDesc,
+        logoUrl: provider.logoUrl,
+        status: provider.status,
+        canUpgrade: provider.canUpgrade,
+        regions: provider.regions,
```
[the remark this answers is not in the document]
pdesoyres-cc: Let's discuss that in sync

src/commands/domain.js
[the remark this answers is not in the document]
davlgd: `domain` could be ok here as it corresponds to the command and how we name it in Console right now. `fqdn` is more a DNS thing, but is technically correct (and should end with a `.` but as we name it clearly, it's ok. Let's sync about it, but imho `fqdn`, while imperfect, is a good choice here

src/commands/profile.js
hsablonniere: Let's go for :

```
id: user.id,
email: user.email,
name: user.name,
avatar: user.avatar,
creationDate: new Date(user.creationDate),
lang: user.lang,
has2FA: (user.preferredMFA != null && user.preferredMFA !== 'NONE'),
```

src/commands/addon.js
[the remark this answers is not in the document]
hsablonniere: Let's go for "we do like the human format" => regions on provider and not on plans.

Let's keep region for now.

hsablonniere (approved):
Impressive work!!! LGTM :+1:

PR #766, opened 2024-06-24
feat(stop): allow to also cancel any ongoing deployment

davlgd:
This PR adds `--cancel` and `-c` to the `stop` command. It allows to ask to stop the app and to cancel any ongoing deployment with only one command.

The branch has been rebased on `support-app-id` from Pierre

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -249,6 +249,9 @@ function run () {
     }),
     exitOnDeploy: getExitOnOption(),
     sameCommitPolicy: getSameCommitPolicyOption(),
+    stopAll: cliparse.flag('all', {
+      description: 'Stop application and any ongoing deployment',
+    }),
     webhookFormat: cliparse.option('format', {
       metavar: 'format',
       default: 'raw',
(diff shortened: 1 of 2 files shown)


hsablonniere (commented):
:+1: I'm OK with the code but I want to discuss the UX...

* a `clever stop --cancel` fails if there is no ongoing deployment
* why no `clever stop && clever cancel-deploy`?
* is `--cancel` clear enough? should it be `clever stop --cancel-deploy`?

davlgd:
> 👍 I'm OK with the code but I want to discuss the UX...
> 
> * a `clever stop --cancel` with fail if there is no ongoing deployment
> * why no `clever stop && clever cancel-deploy`?
> * is `--cancel` clear enough? should it be `clever stop --cancel-deploy`?

A `clever stop && clever cancel-deploy` is what I try to avoid here. And I'm often in the case where I want to stop every deployment of an app, it's why I did this at first. 

For the flag name I'm open. Help is clear enough but the command is `cancel-deploy`. We can also use a more generic `--force` for example.

davlgd:
As discussed in sync it's now: 
* `--all` flag
* Don't `cancel-deploy` if there is no ongoing deployment, without throwing an error
* Rebased on master

aurrelhebert (approved):
LGTM! Looks clean to me

hsablonniere (changes requested):
I don't think a command should import another command. I'll rework the commit myself.

hsablonniere (changes requested):
I updated the code so a command does not rely on another but I'm not sure this update is necessary.

From my tests, a stop on an app cancels any ongoing deployments (new commits and upscaling). Let's do some more tests before merging this...

PR #767, opened 2024-06-24
feat(console): open the Console even if no app is linked

davlgd:
This PR has been rebased on `support-app-id` from Pierre


hsablonniere (approved):
Thanks, let's go for it! :tada: 

FYI, I moved the first line a bit lower as we do this everywhere.

PR #787, opened 2024-07-01
docs(setup-systems): reintroduce nix

### LINKED ISSUE #748: Nix packaging: move to nixpkgs

florian-sanders-cc:
Fixes #748

The `clever-tools` package is now available directly on `nixpkgs - unstable`.

This PR simply updates the docs to reflect that.

## TODO:

- [x] wait for the nixpkgs website to reflect the `clever-tools` availability (right now searching for the package yields nothing even though it's available and it works when adding it to your config)
- [x] create an issue to update the docs once the package makes it to the latest release branch (next major version probably unless it's backported before that but it's unlikely).

### DIFF
--- a/docs/setup-systems.md
+++ b/docs/setup-systems.md
@@ -15,6 +15,7 @@ Clever Cloud CLI is based on Node.js. We thought it to be easily available on an
   - [Chocolatey](/docs/setup-systems.md#chocolatey)
   - [Binary (.zip)](/docs/setup-systems.md#binary-zip)
 - [Docker](/docs/setup-systems.md#docker)
+- [Nix package manager](/docs/setup-systems.md#nix-package-manager)
 
 ## GNU/Linux
 
@@ -144,3 +145,7 @@ In your `Dockerfile` copy `clever-tools` from the image itself with a simple one
 ```Dockerfile
 COPY --from=clevercloud/clever-tools /bin/clever /usr/local/bin/clever
 ```
+
+## Nix package manager
+
+If you are using Nix on NixOS or any other compatible system, the package is available in `unstable` channel. Follow [these instructions](https://search.nixos.org/packages?channel=unstable&show=clever-tools&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Clever Tools
 
-Deploy on Clever Cloud and control you applications, add-ons, services from command line.
+Deploy on Clever Cloud and control your applications, add-ons, services from command line.
 
 - [Create a Clever Cloud account](https://console.clever-cloud.com)
 


docs/setup-systems.md
```diff
@@ -15,6 +15,7 @@ Clever Cloud CLI is based on Node.js. We thought it to be easily available on an
   - [Chocolatey](/docs/setup-systems.md#chocolatey)
   - [Binary (.zip)](/docs/setup-systems.md#binary-zip)
 - [Docker](/docs/setup-systems.md#docker)
+- [Nix](/docs/setup-systems.md#nix-package-manager)
```
davlgd, line 18:
```suggestion
- [Nix package manager](/docs/setup-systems.md#nix-package-manager)
```

docs/setup-systems.md
```diff
@@ -144,3 +145,8 @@ In your `Dockerfile` copy `clever-tools` from the image itself with a simple one
 ```Dockerfile
 COPY --from=clevercloud/clever-tools /bin/clever /usr/local/bin/clever
 ```
+
+### Nix package manager
+
+If you are using Nix or NixOS, the package is available on [nixpkgs - unstable](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
```
davlgd, line 151:
```suggestion
If you are using Nix on NixOS or any other compatible system, the package is available on [nixpkgs - unstable](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
```

docs/setup-systems.md
```diff
@@ -144,3 +145,8 @@ In your `Dockerfile` copy `clever-tools` from the image itself with a simple one
...
+You may add it to your configuration just like any other package.
```
davlgd, line 152:
```suggestion
[Add it](LINK) to your configuration.
```
If there is no specific command for that, replace LINK with a link to Nix doc about it.

davlgd (changes requested):
Thanks for that! Just some suggestions/nitpicks on texts.

docs/setup-systems.md
davlgd: LGTM, my main concern here is to redirect to Nix documentation for new users to know where to find more informations.

hsablonniere:
@florian-sanders-cc Thanks for this. I'll let you and @davlgd settle the details and let me know when we can merge :wink:

florian-sanders-cc:
@davlgd I think it's ready, I've handled your feedback, tell me if there's more :wink:

docs/setup-systems.md
florian-sanders-cc: Found a better solution, the nixpkgs search result shows a "how to install section". It wasn't available at the time of the PR but now it is so I'll link to that:
nixpkgs search result for clever-tools - how to install section

https://search.nixos.org/packages?channel=unstable&show=clever-tools&from=0&size=50&sort=relevance&type=packages&query=clever-tools

docs/setup-systems.md
```diff
@@ -144,3 +145,8 @@ In your `Dockerfile` copy `clever-tools` from the image itself with a simple one
 ```Dockerfile
 COPY --from=clevercloud/clever-tools /bin/clever /usr/local/bin/clever
 ```
+
+## Nix package manager
+
+If you are using Nix on NixOS or any other compatible system, the package is available on [nixpkgs - unstable](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
+To install the package, follow the guildelines from [How to install clever-tools - nixpkgs](https://search.nixos.org/packages?channel=unstable&show=clever-tools&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
```
davlgd, line 152:
```suggestion
If you are using Nix on NixOS or any other compatible system, the package is available in `unstable` channel. Follow [these instructions](https://search.nixos.org/packages?channel=unstable&show=clever-tools&from=0&size=50&sort=relevance&type=packages&query=clever-tools).
```

florian-sanders-cc: perfect, thanks

davlgd (changes requested):
One last nitpick to reduce repetitions/instructions.

florian-sanders-cc:
Better this way indeed, thanks @davlgd should be good now!

davlgd:
LGTM @hsablonniere, you can merge. Thanks @florian-sanders-cc !

hsablonniere (approved):
Yeah :+1: LGTM :wink:

PR #792, opened 2024-07-02
docs: update some texts

davlgd:
Also reorder some options (`--format` always at the end for example)

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -169,7 +169,7 @@ function run () {
       metavar: 'after',
       aliases: ['since'],
       parser: Parsers.date,
-      description: 'Fetch logs after this date/time (ISO8601)',
+      description: 'Fetch logs after this date/time (ISO8601 date, positive number in seconds or duration, e.g.: 1h)',
     }),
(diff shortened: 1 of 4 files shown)


bin/clever.js
```diff
@@ -190,7 +190,7 @@ function run () {
       metavar: 'before',
       aliases: ['until'],
       parser: Parsers.date,
-      description: 'Fetch logs before this date/time (ISO8601 date or duration, positive number in seconds or duration Ex: 1h)',
+      description: 'Fetch logs before this date/time (ISO8601 date, positive number in seconds or duration, e.g.: 1h)',
```
hsablonniere, line 193: suggestion: can you also do this on `after`/`since`?

hsablonniere (changes requested):
Thanks, I added one suggestion.

bin/clever.js
davlgd: It's ok and rebased

hsablonniere (approved):
LGTM :+1:

PR #793, opened 2024-07-02
fix(drain): remove double --addon option

### LINKED ISSUE #786: Drain: double `--addon` option
The `clever drain create` command have a double `--addon` option (one from `clever drain`). We need to use private option mechanism to resolve this.

davlgd:
Fix #786


hsablonniere (approved):
LGTM :+1:

PR #798, opened 2024-07-04
feat(docker): add curl and jq binaries

miton18:
This is the work of @chmuche
https://github.com/CleverCloud/clever-tools-dockerhub/pull/6

but applied on the docker template

* [ ] Remove the last commit which is just for testing purpose

## How to test

* build test image `docker build -t cc:test templates/dockerhub/test`
* start container `docker run --rm -it --entrypoint sh cc:test`
* use `curl` should work: `curl http://example.com`
* use `jq` should work: `echo '{"hello":"jq"}' | jq`

### DIFF
--- a/templates/dockerhub/Dockerfile
+++ b/templates/dockerhub/Dockerfile
@@ -12,7 +12,7 @@ RUN curl --output clever-tools_linux.tar.gz https://clever-tools.clever-cloud.co
 # Only grep the clever-tools binary and his libraries for the release stage
 # We use ldd to find the shared object dependencies.
 RUN \
-	mkdir -p /tmp/fakeroot/lib  && \
+	mkdir -p /tmp/fakeroot/lib && \
 	cp $(ldd /usr/local/bin/clever | grep -o '/.\+\.so[^ ]*' | sort | uniq) /tmp/fakeroot/lib && \
 	for lib in /tmp/fakeroot/lib/*; do strip --strip-all $lib; done && \
 	mkdir -p /tmp/fakeroot/bin/ && \
@@ -28,9 +28,14 @@ LABEL version="<%= version %>" \
 VOLUME ["/actions"]
 WORKDIR /actions
 
-COPY --from=build /tmp/fakeroot/ /
+RUN \
+    ## The loader search ld-linux-x86-64.so.2 in /lib64 but the folder does not exist
+    ln -s lib lib64 && \
+    mkdir -p /etc/ssl/certs
 
-## The loader search ld-linux-x86-64.so.2 in /lib64 but the folder does not exist
-RUN ln -s lib lib64
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=build /tmp/fakeroot/ /
+COPY --from=ghcr.io/tarampampam/curl:8.11.1 /bin/curl /usr/bin/curl
+COPY --from=ghcr.io/jqlang/jq /jq /usr/bin/jq
 
 ENTRYPOINT ["clever"]
--- a/templates/dockerhub/README.md
+++ b/templates/dockerhub/README.md
@@ -1,6 +1,6 @@
 # Clever Tools Docker Image
 
-This is a lightweight Docker image intended to be used mostly in CI environment. 
+This is a lightweight Docker image intended to be used mostly in CI environment.
 
 ## How to use
 


templates/dockerhub/Dockerfile
```diff
@@ -12,11 +12,11 @@ RUN curl --output clever-tools_linux.tar.gz https://clever-tools.clever-cloud.co
 # Only grep the clever-tools binary and his libraries for the release stage
 # We use ldd to find the shared object dependencies.
 RUN \
-	mkdir -p /tmp/fakeroot/lib  && \
+	mkdir -p /tmp/fakeroot/lib && \
 	cp $(ldd /usr/local/bin/clever | grep -o '/.\+\.so[^ ]*' | sort | uniq) /tmp/fakeroot/lib && \
 	for lib in /tmp/fakeroot/lib/*; do strip --strip-all $lib; done && \
 	mkdir -p /tmp/fakeroot/bin/ && \
-	cp /usr/local/bin/clever /tmp/fakeroot/bin/
+	cp /usr/local/bin/clever /tmp/fakeroot/bin/ && \
```
pdesoyres-cc, line 19: fix: Remove the last `&&`

templates/dockerhub/Dockerfile
```diff
@@ -28,9 +28,11 @@ LABEL version="<%= version %>" \
 VOLUME ["/actions"]
 WORKDIR /actions
 
-COPY --from=build /tmp/fakeroot/ /
+RUN mkdir -p /etc/ssl/certs
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
pdesoyres-cc, line 32: question: I don't understand this one: I read that we copy a file at the same place, does this make sens?

templates/dockerhub/Dockerfile
```diff
@@ -28,9 +28,11 @@ LABEL version="<%= version %>" \
...
 
-## The loader search ld-linux-x86-64.so.2 in /lib64 but the folder does not exist
-RUN ln -s lib lib64
+COPY --from=build /tmp/fakeroot/ /
+COPY --from=ghcr.io/tarampampam/curl:8.6.0 /bin/curl /usr/bin/curl
```
pdesoyres-cc, line 35: question: why not use `apt-get` instead?

templates/dockerhub/Dockerfile
```diff
@@ -28,9 +28,11 @@ LABEL version="<%= version %>" \
...
+COPY --from=ghcr.io/jqlang/jq /jq /usr/bin/jq
```
pdesoyres-cc, line 36: question: why not use `apt-get` instead?

templates/dockerhub/test/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM debian AS build
```
pdesoyres-cc, line 1: What is this file for?

pdesoyres-cc (changes requested):
Nice, but with some nitpicks and questions.

templates/dockerhub/Dockerfile
chmuche: Inside the base image, busybox:glibc there is not cert or packet manager. So we copy the certs from the build image with the `--from`

templates/dockerhub/Dockerfile
chmuche: And copy it from the `build` image ?
By the way, I add curl to this image and I can get script or call we hoot inside my CI.

templates/dockerhub/test/Dockerfile
chmuche: Splitting docker image, that way the final image is smaller and without security risk.

chmuche (commented):
I the main author of the PR, but in the wrong repo. I hope my answer are ok.

pdesoyres-cc:
@chmuche we're going back to this PR and we plan to merge it soon. Thanks again for contributing.

I've test the image and it works fine.

I'm wondering why we need the `RUN ln -s lib lib64` instruction and if we still need it now. I'm not sure, we should remove it it this PR (Or at least, it should be in a different commit). @hsablonniere do you have any idea about this?

templates/dockerhub/Dockerfile
hsablonniere: Once we are in the `FROM busybox:glibc AS release`, we're no longer in a debian.

hsablonniere:
Thanks very much @chmuche @miton18 and @pdesoyres-cc for this :+1:

hsablonniere (approved):

Issue #807, opened 2024-07-19
Bundle ESM to CJS before building auto packaged version


hsablonniere:
* [x] bundle to single CJS file with rollup
* [x] add rollup phase to job-build
* [x] enable sourcemaps

Issue #773, opened 2024-06-26
Remove the lazyRequire system


hsablonniere:

hsablonniere:
Fixed by #804

Issue #771, opened 2024-06-26
Migrate the project from CJS to ESM


hsablonniere:
And maybe bundle the code before sending it to pkg...

hsablonniere:
Fixed by #806

PR #809, opened 2024-07-25
Big rework (no lazy require, CJS => ESM sources, rollup to CJS build)

hsablonniere:
This PR includes several major refactors and some late changes fixes for this refactor:

* no lazy require #773
* CJS => ESM sources #771
* rollup to CJS build #807
* Update to new clever-client with breaking change on crypto #830
* Use @clevercloud/client esm instead of cjs #831
* Fix for database default command not working post rework #829
* Use `node:` for Node.js modules


rollup.config.js
```diff
@@ -0,0 +1,76 @@
...
+  output: {
+    file: 'build/clever.cjs',
+    format: 'cjs',
+    sourcemap: 'inline',
+  },
+  plugins: [
+    {
+      transform (code, id) {
+
+        // formiable (used by superagent) hijacks require :-(
```
miton18, line 18: typo ?

src/load-package-json.cjs
miton18, line 1: maybe later https://nodejs.org/dist/v18.15.0/docs/api/esm.html#esm_import_assertions

miton18 (changes requested):

miton18:
QA: Works on my side

rollup.config.js
hsablonniere: Good catch

src/load-package-json.cjs
hsablonniere: Yes, except the syntax and name of the feature has changed, it's now the `with` keyword and Node unshipped the `assert` syntax.

https://nodejs.org/api/esm.html#import-attributes

We'll be able to go for v22 soon.

PR #818, opened 2024-09-04
fix: addon delete without org parameter

### LINKED ISSUE #811: `clever addon delete` need always `--orga` ?
I tried to delete a postgresql addon with `clever addon delete` but I need to force the `--orga`

```
export DATABASE_ID=$(clever curl $CURL_APP_URL/addons | jq '.[] | select(.provider.id == "postgresql-addon") | .id' -r)
clever addon delete $DATABASE_ID --verbose --yes
[DEBUG] Load configuration from environment variables
[DEBUG] Load configuration from /home/apasquier/.config/clever-cloud/clever-tools.json
[DEBUG] DELETE https://api.clever-cloud.com/v2/self/addons/addon_909f7994-30b3-403e-9c11-16d9bf4be363 ? undefined
```

Only `/self`  is used.

The `$DATABASE_ID` is a valid one because `clever addon delete $DATABASE_ID --org=$ORGA_ID --yes` works

davlgd:
This PR: 

- add a `findByName` function in add-ons model, to find an add-on only by its name
- look for orgId by add-on ID or name if none is specified

Fixes #811

### DIFF
--- a/src/commands/addon.js
+++ b/src/commands/addon.js
@@ -176,7 +176,17 @@ async function deleteAddon (params) {
   const { yes: skipConfirmation, org: orgaIdOrName } = params.options;
   const [addon] = params.args;
 
-  const ownerId = await Organisation.getId(orgaIdOrName);
+  let ownerId = await Organisation.getId(orgaIdOrName);
+
+  if (ownerId == null && addon.addon_id != null) {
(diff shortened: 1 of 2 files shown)


aurrelhebert (approved):
QA OK, run a couple of tests: 
 - [x] Tests add-ons in user, orga
 - [x] Tests delete with owner or yes option
 - [ ] Tests add-ons list command

PR #821, opened 2024-09-06
fix(update-notifier): set `isGlobal` to `true`

### LINKED ISSUE #812: update-notifier: use the `isGlobal` flag to improve the notification message
## Context

When working with @HeleneAmouzou, we noticed that the update message instruction was to run `npm i` instead of `npm i -g`.

It's a low impact / low prio issue but also a quick win since @davlgd pointed out there was a flag to make `update-notifier` show the proper message.

TLDR; we should set the `isGlobal` flag to `true` so the message is correct. 

I'll create a PR for this when I have time (soon)

florian-sanders-cc:
Fixes #812

## What does this PR do?

This PR enables the `isGlobal` flag of the `update-notifier` as discussed with @davlgd.

The update message now rightfully specifies to use `npm i -g` instead of `npm i`
usual clever tools update message specifying to use npm i -g clever-tools

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -43,7 +43,7 @@ if (!updateNotifierExplicitFalse && !isRunThroughPackagedBinary) {
   }).notify({
     isGlobal: true,
     getDetails () {
-      const docsUrl = 'https://www.clever-cloud.com/doc/clever-tools/getting_started';
+      const docsUrl = 'https://github.com/CleverCloud/clever-tools/tree/master/docs#how-to-use-clever-tools';
       return `\nPlease follow this link to update your clever-tools:\n${docsUrl}`;
     },
   });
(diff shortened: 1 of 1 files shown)


bin/clever.js
```diff
@@ -41,6 +41,7 @@ if (!updateNotifierExplicitFalse && !isRunThroughPackagedBinary) {
     pkg,
     tagsUrl: 'https://api.github.com/repos/CleverCloud/clever-tools/tags',
   }).notify({
+    isGlobal: true,
     getDetails () {
       const docsUrl = 'https://www.clever-cloud.com/doc/clever-tools/getting_started';
```
davlgd, line 46:
```suggestion
      const docsUrl = 'https://github.com/CleverCloud/clever-tools/tree/master/docs#how-to-use-clever-tools';
```

davlgd (changes requested):
LGTM, juste an URL change to add. 

To check it I edited the `package.json` version and added `updateCheckInterval: 0,` in `updateNotifierModule.notify()` as explained [here](https://github.com/yeoman/update-notifier/blob/HEAD/example.js)

Thanks @florian-sanders-cc for the link ^^

miton18 (approved):

hsablonniere (approved):
Thank you @florian-sanders-cc for this and @davlgd @miton18 for the reviews!

PR #827, opened 2024-10-01
feat(domain): init diag command

florian-sanders-cc:
## What does this PR do?

- Adds a `clever domain diag [--app {appId|appName}] [--domain {string}] [-F json]` command,
- This command allows users to get diagnostics about DNS configs for every domain associated to a given app,

Some important context:

- we throw errors and tell users their DNS config is wrong when:
  - no config was found (no CNAME / no A records),
  - the config contains at least one unknown record (a record that does not match the default Load Balancer config of the app).
- APEX domains can only be configured with A records so diagnostics for such domains will never mention CNAME records,
- we don't support configs relying on a CDN (like Cloudflare for instance) and we don't support private Load Balancers for now (no API).

## How to review?

- You should probably review the PR from the client first: https://github.com/CleverCloud/clever-client.js/pull/116
- Check the last commit, ignore the `chore` commit (this part should be reviewed within the client project),
- Play with the command :+1:
`clever domain diag [--app {appId|appName}] [--domain {string}] [-F json]` 

## TODO

- [x] rebase on big rework once it's merged on master
- [x] replace local client with new published version once it's released

### DIFF
--- a/src/commands/domain.js
+++ b/src/commands/domain.js
@@ -2,6 +2,16 @@ import * as Application from '../models/application.js';
 import { Logger } from '../logger.js';
 import { get as getApp, addDomain, getFavouriteDomain as getFavouriteDomainWithError, markFavouriteDomain, unmarkFavouriteDomain, removeDomain } from '@clevercloud/client/esm/api/v2/application.js';
 import { sendToApi } from '../models/send-to-api.js';
+import colors from 'colors/safe.js';
+import { parse as parseDomain } from 'tldts';
+import { diagDomainConfig } from '@clevercloud/client/esm/utils/diag-domain-config.js';
+import { DnsResolver } from '../models/node-dns-resolver.js';
+
+/**
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').DomainInfo} DomainInfo
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').ResolveDnsResult} ResolveDnsResult
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').DomainDiag} DomainDiag
+ */
 
 function getFavouriteDomain ({ ownerId, appId }) {
   return getFavouriteDomainWithError({ id: ownerId, appId })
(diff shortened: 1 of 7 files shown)


src/commands/domain.js
```diff
@@ -1,16 +1,19 @@
...
+import { diagDomainConfig } from '@clevercloud/client/esm/utils/diag-domain-config.js';
+import { NodeDnsResolver } from '../models/node-dns-resolver.js';
+
+/**
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').DomainInfo} DomainInfo
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').ResolveDnsResult} ResolveDnsResult
+ * @typedef {import('@clevercloud/client/esm/utils/diag-domain-config.types.js').DomainDiag} DomainDiag
+ */
+
+const resolver = new NodeDnsResolver();
```
pdesoyres-cc, line 16: question: do we really want to instanciate the resolver here? Shouldn't we instantiate a resolver only when we need it?

src/commands/domain.js
```diff
@@ -89,4 +92,206 @@ async function rm (params) {
...
+    }
+  }
+
+  switch (format) {
+    case 'json':
+      Logger.printJson(allDomainDiagnostics);
+      break;
+    default: {
+      if (allDomainDiagnostics.length > 0) {
+        Logger.println(colors.yellow('\n⚠ TODO avec David, une phrase pour dire qu\'on ne supporte pas tout + se référer à la doc de la commande + un lien vers feedback ?'));
```
pdesoyres-cc, line 134: nitpick: do not forget this todo

src/commands/domain.js
```diff
@@ -89,4 +92,206 @@ async function rm (params) {
...
+  const unknownCname = diagDetails.find((diag) => diag.code === 'unknown-cname');
+
+  const hasARecords = resolvedDnsConfig.aRecords.length > 0;
+  const hasCnameRecord = resolvedDnsConfig.cnameRecords.length > 0;
+
+  Logger.println('\n' + colors.blue(hostname) + ' ' + colors.yellow(pathPrefix) + '\n');
+
+  switch (diagSummary) {
+    case 'managed':
+      Logger.println(' '.repeat(2) + colors.green('✔ Managed by Clever Cloud'));
```
pdesoyres-cc, line 171: suggestion: extract `' '.repeat(2)` into a const `const indent2 = ' '.repeat(2);`, and use it everywhere below.

src/commands/domain.js
```diff
@@ -89,4 +92,206 @@ async function rm (params) {
...
+    validDiags.forEach((diag) => {
+      Logger.println(' '.repeat(2) + diag.record.value, colors.green('✔ A Record OK'), hasCnameRecord ? `(from CNAME ${resolvedDnsConfig.cnameRecords[0]}.)` : '');
+    });
+  }
+
+  // Replace A with CNAME
+  if (diagSummary === 'valid' && suggestedCname != null) {
+    Logger.println('');
+    Logger.println(' '.repeat(2) + 'ⓘ You can replace your A records with this CNAME:');
+    Logger.println(' '.repeat(2) + ' '.repeat(4) + suggestedCname.record.value);
```
pdesoyres-cc, line 200: suggestion: extract `' '.repeat(4)` into a const `const indent4 = ' '.repeat(4);`, and use it everywhere below.
Or use a `const indent6 = ' '.repeat(6);` and replace all `' '.repeat(2) + ' '.repeat(4)`.

src/commands/domain.js
```diff
@@ -89,4 +92,206 @@ async function rm (params) {
...
+  }
+
+  switch (format) {
+    case 'json':
+      Logger.printJson(allDomainDiagnostics);
+      break;
+    default: {
+      if (allDomainDiagnostics.length > 0) {
+        Logger.println(colors.yellow('\n⚠ TODO avec David, une phrase pour dire qu\'on ne supporte pas tout + se référer à la doc de la commande + un lien vers feedback ?'));
+
```
pdesoyres-cc, line 135: nitpick: extra space

src/commands/domain.js
pdesoyres-cc: You could also define a function that prints indented text:
```js
function println(text, indent = 0) {
  Logger.println(`${' '.repeat(indent)}${text}`);
}
```

pdesoyres-cc (approved):
Really nice work Florian. Well done.

I have juste some nitpicks and suggestions.

src/models/node-dns-resolver.js
```diff
@@ -0,0 +1,43 @@
+import { Resolver } from 'node:dns/promises';
+
+export class NodeDnsResolver {
```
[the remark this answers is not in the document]
florian-sanders-cc: There used to be both a `DnsOverHttpsResolver` and `NodeDnsResolver` but since we've removed the DoH one, we can rename this one, nice catch :wink:

src/commands/domain.js
florian-sanders-cc: Discussed with David & Hubert, we'll remove this warning, these will be included in docs but not the command itself.

src/commands/domain.js
florian-sanders-cc: Eh yeah, `' '.repeat(2) + ' '.repeat(4)` is silly :sweat_smile: 

I'll go with the function, thanks for the suggestion :+1:

pdesoyres-cc (approved):
LGTM

src/models/node-dns-resolver.js
```diff
@@ -19,7 +19,7 @@ export class NodeDnsResolver {
         case 'ENODATA':
           return [];
       }
-      console.log('error', error);
+      throw new Error(error);
```
pdesoyres-cc, line 22: `err` is already an instance of `Error` so you should not create a new `Error` object around it.

Note that, when throwing an `Error`, the `Error#message` is printed to the console. Here, the raw error coming from `node:dns` may be cryptic for end user. So, maybe you want to tell why this error occurs:

```suggestion
      throw new Error(`Could not resolve DNS for ${hostname}. Caused by: ${error.message}`);
```

src/models/node-dns-resolver.js
```diff
@@ -37,7 +37,7 @@ export class NodeDnsResolver {
         case 'ENODATA':
           return null;
       }
-      console.log('error', error);
+      throw new Error(error);
```
pdesoyres-cc, line 40: same as above

pdesoyres-cc (changes requested):
I left a comment about error handling.

src/models/node-dns-resolver.js
florian-sanders-cc: I rushed my modifications here :see_no_evil:, nice suggestion, I've applied it :+1:

pdesoyres-cc (approved):
LGTM. Well done Florian

hsablonniere (approved):
LGTM, well done @florian-sanders-cc for this work :clap: :+1: and thanks again @davlgd for your tests, inputs, feedbacks and fixes :heart:

PR #833, opened 2024-10-16
fix(addon): update Beta providers list

davlgd:
I've added Matomo as beta, its current status, and declared KC/Metabase as beta after their update. This commit was based on refactor branch


hsablonniere (approved):
LGTM :+1:

PR #834, opened 2024-10-16
feat(addon): add URL to use/manage add-on at creation

davlgd:
For add-ons with a public management URL, we tell the user what the URL is at creation, to ease onboarding. For that I added add-on env to returned object after creation (we'll move this to API side in a future iteration). 

I've only done this for KC, Matomo, Metabase for now, but we'll can go further for other add-ons too. This commit was based on refactor branch


hsablonniere (approved):
LGTM, well done @davlgd :+1:

PR #839, opened 2024-10-17
refactor(domains/diagApplication): rely on `sortDomains` from the client

florian-sanders-cc:
## What does this PR do?

- Removes the `sortDomains` function from the `domains.js` file,
- Use the `sortDomains` callback provided by the client project (this way the sort order matches the one used within the component project). 

## Should this be part of 3.9.0?

Not essential but it would make sense.
If it's not part of the 3.9.0 it's perfectly fine, it's just a tiny refactor.


hsablonniere (approved):
LGTM :+1:

PR #842, opened 2024-10-22
chore: add token expiry to clever profile

### LINKED ISSUE #668: Auth expiration
Tokens crafted for the CLI will soon have a life time of 12 months.
Maybe we can take care of it and:

* Store the expiration date in `.clever-tool.json`
* Print a warning messages some days before the expiration
* Check if an errored call to the API result in a unretryable action

Galimede:
Fixes #668 

_Note: date format is the same as the one used in the console_

### DIFF
--- a/src/commands/profile.js
+++ b/src/commands/profile.js
@@ -7,6 +7,17 @@ export async function profile (params) {
   const { format } = params.options;
 
   const user = await User.getCurrent();
+  const currentToken = await User.getCurrentToken();
+  const tokenExpiration = new Date(currentToken.expirationDate).toLocaleString('en-US', {
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+    hour: 'numeric',
+    minute: 'numeric',
+    hour12: false,
+    timeZone: 'UTC',
+    timeZoneName: 'short',
+  });
 
   const formattedUser = {
     id: user.id,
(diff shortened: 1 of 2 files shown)


davlgd (commented):
Thanks for this! Just two nitpicks on date/time formatting, I let @hsablonniere tell what's his preference on this.

hsablonniere (changes requested):
Thanks @Galimede, well done :+1: 

About the date formatting:

* I looked at the console in the tokens page
  * As @Galimede said, he applied the same format
  * Yet, the console has a mention for UTC in the table header
* I also looked at other dates displayed in the CLI
  * The rare cases (backups, activity) are in ISO

For this command, I want to follow the console and display a human date (no ISO).

* Let's add the `timeZoneName: 'short'`
* I would be for keeping the AM/PM hour display as it's more custom to english locales but I'm also OK with `hour12: false`
* I'd like to keep the `day: 'numeric'` as it's more human
* Why not using `hour: '2-digit'` but FYI, `hour12: true` forces the hours to have 2 digits
* `minute: '2-digit' doesn't change anything
* FYI, UTC in `timeZone: 'UTC' should be kept uppercase

hsablonniere:
@Galimede FYI, We discussed and applied changes during a meeting with @davlgd and @aurrelhebert

PR #846, opened 2024-11-05
fix(domain): add missing spacing within diag report

florian-sanders-cc:
## What does this PR do?

- Adds a missing space between the record and the text "from (...)" when there are invalid A records coming from a CNAME

In prod:
image

With the fix:
image


hsablonniere (approved):
LGTM, tested on built version! :+1:

PR #851, opened 2024-11-21
feat(profile): open in the Console

davlgd:
This PR allow an user to open his user profile in the console with the command `clever profile open`. I also make texts the same for such commands.


hsablonniere (approved):
LGTM :+1:

PR #852, opened 2024-11-21
refactor: reorganize config variables and declare new ones for common URLs

davlgd:
As discussed previously, this PR defines a `const` with URLs we need in multiple parts of the CLI to define them once, use them when needed and make update easier.


hsablonniere (approved):
LGTM :+1

PR #853, opened 2024-11-21
feat(config): add tasks support

davlgd:
This PR add `task` as an option to enable/disable with `REGULAR` or `TASK` as accepted values in the `config` command. It allows to manage it once a task app is created.

The "trick" here, it that's it's not a config option for API, but it's the better place to handle it for now and we do the same in Console.

EDIT HUBERT:

I rewrote the PR to improve the existing code (not related to the PR) before adding the support for `task`. I kept most of what David proposed (better display in table, better help...) and added a few things (fix error handling or argument/flag behaviour).

### DIFF
--- a/src/commands/config.js
+++ b/src/commands/config.js
@@ -1,7 +1,5 @@
-import { update as updateApplication } from '@clevercloud/client/esm/api/v2/application.js';
 import * as Application from '../models/application.js';
 import * as ApplicationConfiguration from '../models/application_configuration.js';
-import { sendToApi } from '../models/send-to-api.js';
 import { Logger } from '../logger.js';
 import colors from 'colors/safe.js';
 
@@ -28,7 +26,7 @@ export async function set (params) {
   const options = {
     [config.name]: ApplicationConfiguration.parse(config, configurationValue),
   };
-  const app = await updateApplication({ id: ownerId, appId }, options).then(sendToApi);
+  const app = await Application.updateOptions(ownerId, appId, options);
   Logger.printSuccess(`Config ${colors.green(config.id)} successfully updated to ${colors.green(ApplicationConfiguration.formatValue(config, app[config.name]))}!`);
 }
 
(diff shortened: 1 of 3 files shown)


src/commands/config.js
```diff
@@ -1,35 +1,33 @@
...
-  if (config != null) {
-    const app = await updateApplication({ id: ownerId, appId }, { [config.name]: ApplicationConfiguration.parse(config, configurationValue) }).then(sendToApi);
-
-    ApplicationConfiguration.printById(app, configurationName);
-  }
+  const patchedApp = {
+    [config.name]: ApplicationConfiguration.parse(config, configurationValue),
+  };
+  const app = await Application.updateOptions(ownerId, appId, patchedApp);
+  Logger.printSuccess(`Config ${colors.grey(config.id)} successfully updated to ${colors.grey(ApplicationConfiguration.formatValue(config, app[config.name]))}!`);
```
davlgd, line 30: On success message I previously used same color as the icon for text to highlight

```suggestion
  Logger.printSuccess(`Config ${colors.green(config.id)} successfully updated to ${colors.green(ApplicationConfiguration.formatValue(config, app[config.name]))}!`);
```

src/commands/config.js
```diff
@@ -1,35 +1,33 @@
...
   const { alias, app: appIdOrName } = params.options;
   const { ownerId, appId } = await Application.resolveId(appIdOrName, alias);
   const config = ApplicationConfiguration.getById(configurationName);
-
-  if (config != null) {
-    const app = await updateApplication({ id: ownerId, appId }, { [config.name]: ApplicationConfiguration.parse(config, configurationValue) }).then(sendToApi);
-
-    ApplicationConfiguration.printById(app, configurationName);
-  }
+  const patchedApp = {
```
davlgd, line 26: as we normalize to true/false for options, maybe we can check the option value with a parser or something to refuse invalid values (here if I set `task` to `titi` I get a successfully set to `true`).

PR #854, opened 2024-11-26
feat(addon): elastic plugins option support

davlgd:
This PR allows the support of elastic plug-ins at creation in Clever Tools. As this option is a comma-separated list, I separate options from pairs, looking for a `,` followed by a word and a `=` and matching it. 

It also adds a parser to add-ons options. It will help us to clean all this in a future release. Here I stay minimal on change to lower impact. 

I've tested it with multiple options, single options, comma-separated options, no options, and it seems working fine.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -342,6 +342,7 @@ function run () {
     }),
     addonOptions: cliparse.option('option', {
       metavar: 'option',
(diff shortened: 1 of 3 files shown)


aurrelhebert (approved):
PR tested, LGTM :heavy_check_mark:

PR #862, opened 2024-12-06
Beautify apps commands

davlgd:
This PR tries to enhance the way information are printed to the user in application management, adding instructions/tips/link when useful. Feel free to test it, discuss choices made.

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -27,10 +27,6 @@ export async function deploy (params) {
 
   await git.addRemote(appData.alias, appData.deployUrl);
 
-  Logger.println(colors.bold.blue(`Remote application is app_id=${appId}, alias=${appData.alias}, name=${appData.name}`));
-
-  Logger.println(colors.bold.blue(`Remote application belongs to ${ownerId}`));
-
   if (commitIdToPush === remoteHeadCommitId) {
     switch (sameCommitPolicy) {
       case 'ignore':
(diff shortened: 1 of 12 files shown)


src/models/git.js
```diff
@@ -130,3 +131,39 @@ export async function isShallow () {
...
 }
+
+/**
+ * Check if the current directory is a git repository
+ * @returns {boolean}
+ * @public
+ */
+export function isGitRepo () {
+  try {
+    childProcess.execSync('git rev-parse --is-inside-work-tree', {
```
pdesoyres-cc, line 142: So we start assuming user has got `git` binary on its machine.

This is something we need to document. I feel like it is a breaking change right?

src/commands/deploy.js
```diff
@@ -103,3 +75,53 @@ async function getBranchToDeploy (branchName, tagName) {
...
+  Logger.println();
+
+  Logger.println(colors.bold('🔄 Deployment progress'));
+  Logger.println(`   ${colors.blue('→ Pushing source code to Clever Cloud…')}`);
+
+  await git.push(deployUrl, commitIdToPush, force)
+    .catch(async (e) => {
+      const isShallow = await git.isShallow();
+      if (isShallow) {
+        throw new Error('Failed to push your source code because your repository is shallow and therefore cannot be pushed to the Clever Cloud remote.');
```
pdesoyres-cc, line 119: I would be interested in the reason of this.

pdesoyres-cc (approved):
Nice, it's really a better output :+1.
I only looked at the code and LGTM with just a few questions.

src/models/git.js
davlgd: I've made it as is, we can also check if git is here, if not, don't print anything related. But as the user need a git repo with commits, I think it's could be better to print a warning if no git is found. 

I think we already tell in doc git is mandatory, but we can insist more on this.

src/commands/deploy.js
davlgd: It was there before. My point here was mainly to change how things are printed, not how the CLI works (except to add some info about git repo): https://github.com/CleverCloud/clever-tools/blob/fed085e2ba0339f55e966d7c8c6439d4dac71164/src/commands/deploy.js#L73

aurrelhebert:
Quick usage feedback, on the create command, the next phrase is always printed (even on an already git project folder) : 
```
  Next steps:
  ! Commit your changes first:
    $ git add .
    $ git commit -m "Initial commit"
```
This seems logical (with a change of the commit message to "add .clever.json file") as a new `.clever.json` file is added. However even in the case that `.clever.json` file is ignored (and the git project is up to date), those steps are printed

davlgd:
> we loose the success information previously plotted 

It was deleted on purpose, we show it in logs so it was a duplicate information

If've fixes other points

PR #865, opened 2024-12-11
fix(rollup): prevent build failure on MacOS

### LINKED ISSUE #864: Rollup build fails when running on MacOS
## Context

When running `npm run build` on MacOS, you get the following error:

```sh
bin/clever.js → build/clever.cjs...
[!] (plugin commonjs--resolver) RollupError: node_modules/fsevents/fsevents.node (1:0): Unexpected character '�' (Note that you need plugins to import files that are not JavaScript)
node_modules/fsevents/fsevents.node (1:0)
1: ����@�
         @��...
   ^
2:
3:
   *
    h���/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFound...
RollupError: Unexpected character '�'
```

This happens because:
- the whole `curlconverter` module is imported,
- the `curlconverter` codebase relies on `nunjucks` to handle `ansible`, `yaml`, etc. but not `JSON` (and I believe that's the only one we actually use),
- `nunjucks` relies on `chokidar`,
- when building on MacOS, `chokidar` relies on the `fsevents` package which is essentially a binary file (`fsevents.node`) built specifically for `darwin` :skull: ,
- our rollup build tries to parse the binary code from `fsevent.node`,
- it fails and crashes.

## Why is it a problem?

- MacOS contributors cannot build the project (not a big deal since for now, the build step is purely optional),
- The `nix` package relies on the build step so that we don't ship all `node_modules` with the package, only the bundled code. This means the `nix` package fails to build on MacOS at the moment (not a big deal but annoying).

We could modify the `nix` package to handle this case but this is not a `nix` related issue so I'd rather fix it in the CLI.

## How to solve the issue?

- The quick and easy solution is to set the `fsevents` dependency as `external` in the rollup config so that it's not part of the bundle. 
  - This can be tested easily on MacOS by doing one of the following:
    - in `rollup.config.js`, `defineConfig({ ..., external: ['fsevents'], plugins: [ ... ] })`
    - or if run `node node_modules/rollup/dist/bin/rollup -c rollup.config.js -e fsevents`.

- This assumes that `fsevents` is indeed irrelevant for our usage of `curlconverter` which I think is the case. Testing the `clever curl` commands with the built code seems to work just fine :+1: 

- Another option would be to tell rollup to simply copy the `fsevents.node` file instead of trying to parse it. This means we'd have to create a small plugin for that. It shouldn't require a lot of code but it doesn't feel right to create a plugin to include some code that should not be part of the bundle anyway?

Thanks to Maël for pointing out the build failure on MacOS.

florian-sanders-cc:
Fixes #864

## What does this PR do?

- Fixes the `npm run build` command on MacOS.

The rollup build fails because:
- the whole `curlconverter` module is imported,
- the `curlconverter` codebase relies on `nunjucks` to handle `ansible`, `yaml`, etc. but not `JSON` (and I believe that's the only one we actually use),
- `nunjucks` relies on `chokidar`,
- when building on MacOS, `chokidar` relies on the `fsevents` package which is essentially a binary file (`fsevents.node`) built specifically for `darwin` :skull: ,
- our rollup build tries to parse the binary code from `fsevent.node`,
- it fails and crashes.

This is why this PR excludes `fsevents` from the bundle.
We could exclude nunjucks or chokidar but I went for the safest solution because I don't know much about nunjucks or chokidar (although chokidar is meant to watch files so I don't see why we would need that in the bundle).

As explained in the issue, we could also create a plugin to handle `fsevents` properly but since we don't actually need this dependency, I don't think it's a good idea to go there.

## How to test?

- If you're on MacOS, run `npm run build`, it should work (and it shouldn't work on master),
- If you're on MacOS, test the `node build/clever.cjs curl` command (I don't see any other potential impact).

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -11,6 +11,10 @@ export default defineConfig({
     format: 'cjs',
     sourcemap: 'inline',
   },
+  // This dependency is only pulled in when building on MacOS (see https://github.com/CleverCloud/clever-tools/issues/864)
+  // It's a binary file and it shouldn't be parsed by rollup, only copied.
+  // We exclude it from the bundle because it's a dependency that is pulled by `curlconverter` but we don't actually need it.
+  external: ['fsevents'],
   plugins: [
     {
       transform (code, id) {


davlgd (approved):
LGTM, good catch nixers, thanks for this ;)

Build seems ok here, I only get 2 warnings:

```
npm run build

> clever-tools@3.10.1 build
> rollup -c rollup.config.js


bin/clever.js → build/clever.cjs...
(!) Circular dependencies
node_modules/semver/classes/comparator.js -> node_modules/semver/classes/range.js -> node_modules/semver/classes/comparator.js
node_modules/superagent/lib/node/index.js -> node_modules/superagent/lib/node/agent.js -> node_modules/superagent/lib/node/index.js
created build/clever.cjs in 3.1s
```

`curl` command is working with the resulting `clever.cjs`

Mudada:
Same as David so looks good to me !

Thanks for the quick fix fellow rabbit :)

hsablonniere:
Thanks @florian-sanders-cc for this. This is a sad situation but the fix is legit. I'm looking forward to being able to update curlconverter (or event ditch it thanks to API tokens).

hsablonniere (approved):
LGTM :+1:

PR #873, opened 2025-01-25
feat: add ssh-keys command

davlgd:
This PR adds a `ssh-keys` command, to add, remove(-all), list SSH keys from a user account and to open the SSH Keys panel from the Console. 

If no SSH keys are found, instructions to create one are printed.

## TODO

* [x] This PR is based on #874 (don't forget to merge it before this one)

### DIFF
--- a/src/commands/ssh-keys.js
+++ b/src/commands/ssh-keys.js
@@ -1,0 +1,159 @@
+import { confirm } from '../models/interact.js';
+import colors from 'colors/safe.js';
+import { Logger } from '../logger.js';
+import fs from 'node:fs';
+import { sendToApi } from '../models/send-to-api.js';
+import { openBrowser } from '../models/utils.js';
+import dedent from 'dedent';
+import {
+  todo_addSshKey as addSshKey,
+  todo_getSshKeys as getSshKeys,
+  todo_removeSshKey as removeSshKey,
+} from '@clevercloud/client/esm/api/v2/user.js';
+
+/**
+ * List SSH keys of the current user
+ * @param {object} params The command parameters
+ * @param {string} params.options.format The output format
+ */
+export async function list (params) {
+  const { format } = params.options;
+
+  const keys = await getUserSshKeys();
+
+  switch (format) {
+    case 'json': {
+      Logger.printJson(keys);
+      break;
+    }
+    case 'human':
+    default: {
+      if (keys.length === 0) {
+        Logger.println(dedent`
+          ${colors.blue('🔐 No SSH keys')}
+          
+          To list the SSH keys on your local system, use the following command:
+          ${colors.grey('ssh-add -l -E sha256')}
+          
+          To create a new key pair, use the following command:
+          ${colors.grey('ssh-keygen -t ed25519 -f ~/.ssh/id_ed25519_clever -C "An optional comment"')}
+          
+          Then add the public key to your Clever Cloud account:
+          ${colors.grey('clever ssh-keys add myNewKey ~/.ssh/id_ed25519_clever.pub')}
+        `);
+        return;
+      }
+
+      Logger.println(`🔐 ${keys.length} SSH key(s):`);
+      keys.forEach((key) => {
+        Logger.println(` • ${colors.blue(key.name)}`, colors.grey(`(${key.fingerprint})`));
+      });
+    }
+  }
+}
+
+/**
+ * Add a SSH key to the current user
+ * @param {object} params The command parameters
+ * @param {Array<string>} params.args
+ */
+export async function add (params) {
+  const [keyName, filePath] = params.args;
+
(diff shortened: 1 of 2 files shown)


src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+    Logger.println('To create a new key pair, you can use the following command:');
+    Logger.println(colors.grey('ssh-keygen -t ed25519 -f ~/.ssh/id_ed25519_clever -C "An optional comment"'));
+    Logger.println();
+    Logger.println('Then add the public key to your Clever Cloud account:');
+    Logger.println(colors.grey('clever ssh-keys add myNewKey ~/.ssh/id_ed25519_clever.pub'));
+    return;
+  }
+
+  if (keys.length === 1) {
+    Logger.println(`🔐 Current user have 1 SSH key: ${keys[0].name} (${keys[0].fingerprint})`);
```
pdesoyres-cc, line 98:
```suggestion
    Logger.println(`🔐 Current user has 1 SSH key: ${keys[0].name} (${keys[0].fingerprint})`);
```

src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+    return;
+  }
+
+  Logger.println(`🔐 Current user have ${keys.length} SSH keys:`);
```
pdesoyres-cc, line 102:
```suggestion
  Logger.println(`🔐 Current user has ${keys.length} SSH keys:`);
```

src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+  keys.forEach((k) => {
+    Logger.println(` ▶ ${colors.blue(k.name)}`, colors.grey(`(${k.fingerprint})`));
```
pdesoyres-cc, line 104: nitpick: we should always use the same bullet point (this is not the same as the one used in the NG PR)

src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+}
+
+/**
+ * Open the SSH keys management page of the Console in your browser
+ * @returns {Promise<void>} A promise that resolves when the page is opened
+ */
+export async function openConsole () {
```
pdesoyres-cc, line 112: nitpick: you can remove the `async` and return the `openBrowser` promise instead.

src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+
+  if (!existsSync(filePath)) {
+    Logger.error(`File ${filePath} does not exist`);
+    return;
+  }
+
+  const fileContent = readFileSync(filePath, 'utf8').trim();
+  Logger.debug(`SSH key file content: ${fileContent}`);
+
+  executeCommand(addKey, [keyName, fileContent]);
```
pdesoyres-cc, line 45: nitpick: you should either `await` or `return` the `executeCommand` promise

src/commands/ssh-keys.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/** Remove a SSH key from the current user
+ * @param {{ namedArgs: { "ssh-key-name": string } }} params The command parameters
+ */
+export async function remove (params) {
+  const { 'ssh-key-name': keyName } = params.namedArgs;
+  executeCommand(removeKey, [keyName]);
```
pdesoyres-cc, line 53: nitpick: you should either `await` or `return` the `executeCommand` promise

pdesoyres-cc, line 53: oh noo! v2 API sucks:
* it fails with an error 500 when trying to delete an unknown ssh-key.
* and there is no GET or HEAD http method available to get a single ssh-key neither

Maybe we should verify if it exists by using an `await getKeys()`

src/commands/env.js
```diff
@@ -8,8 +8,8 @@ import * as application from '@clevercloud/client/esm/api/v2/application.js';
 
 export async function list (params) {
   const { alias, app: appIdOrName, 'add-export': addExportsOption, format } = params.options;
+  console.log(params.options);
```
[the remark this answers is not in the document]
davlgd: 😬

pdesoyres-cc (approved):

hsablonniere:
Just like I did for #874, I rewrote the commits of this PR to adjust the code.

* same modifications as the emails commands
  * use clever-client (with todo)
  * merge model into commands
* simplify the list display (no more case for just one, better to have colors)
* use big bloc of texts with dedent
* added an example to list local keys `ssh-add -l -E sha256`
* it's a bit weird to have `remove-all` here and `clear` in emails

pdesoyres-cc (approved):
LGTM. I just left a nitpick

PR #874, opened 2025-01-25
feat: add emails command

davlgd:
This PR adds an `emails` commands to add, remove(-all) secondary emails to a user account, set the primary email, list emails, and open the emails management page in the Console. 

I've added an email parser to validate emails. It uses a simple regex to avoid corner cases. There is another proposition in comment, chose the one you prefer. 

This PR includes a new utils to open a browser page.

### DIFF
--- a/src/commands/emails.js
+++ b/src/commands/emails.js
@@ -1,0 +1,174 @@
+import colors from 'colors/safe.js';
+import * as User from '../models/user.js';
+import { Logger } from '../logger.js';
+import { openBrowser } from '../models/utils.js';
+import { sendToApi } from '../models/send-to-api.js';
+import { confirm } from '../models/interact.js';
+import {
+  todo_addEmailAddress as addEmailAddress,
+  todo_getEmailAddresses as getEmailAddresses,
+  todo_removeEmailAddress as removeEmailAddress,
+} from '@clevercloud/client/esm/api/v2/user.js';
+
+/**
+ * Show primary and secondary email addresses of the current user
+ * @param {object} params The command parameters
+ * @param {string} params.options.format The output format
+ */
+export async function list (params) {
+  const { format } = params.options;
+  const addresses = await getUserEmailAddresses();
+
+  switch (format) {
+    case 'json': {
+      Logger.printJson(addresses);
+      break;
+    }
+    case 'human':
+    default: {
+      Logger.println('✉️  Primary email address:');
+      Logger.println(` • ${colors.green(addresses.primary)}`);
+      if (addresses.secondary.length > 0) {
+        Logger.println();
+        Logger.println(`✉️  ${addresses.secondary.length} secondary email address(es):`);
+        addresses.secondary.forEach((address) =>
+          Logger.println(` • ${colors.blue(address)}`),
+        );
+      }
+    }
+  }
+}
+
+/**
+ * Add a secondary email address to the current user
+ * @param {object} params The command parameters
+ * @param {Array<string>} params.args
+ */
+export async function addSecondary (params) {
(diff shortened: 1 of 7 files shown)


src/commands/emails.js
```diff
@@ -0,0 +1,107 @@
...
+      Logger.println(` ▷ ${colors.blue(email)}`),
+    );
+  }
+}
+
+/**
+ * Open the emails management page of the Console in your browser
+ * @returns {Promise<void>} A promise that resolves when the page is opened
+ */
+export async function openConsole () {
```
pdesoyres-cc, line 105: nitpick: remove the `async` and return the `openBrowser` result

src/models/emails-api.js
```diff
@@ -0,0 +1,48 @@
+
+/**
+* GET /self/emails
+* @param {Object} params
```
pdesoyres-cc, line 4: fix: remove this `@param`

pdesoyres-cc (changes requested):
Well done David. I have several comments and some fixes to apply (some of them are due to API limitations or bugs).

src/parsers.js
```diff
@@ -56,6 +56,17 @@ export function date (dateString) {
   return duration;
 }
 
+// Could be:
+// /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
+const emailRegex = /^[^@]+@[^@]+\.[^@]+$/;
```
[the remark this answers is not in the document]
pdesoyres-cc: Anyway, the API has also got an email address validation process and returns a proper error code for that. Maybe we can rely on API validation only?

src/commands/emails.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+/**
+ * Add a secondary email to the current user
+ * @param {object} params The command parameters
+ * @param {string} params.namedArgs.email The email to add
+ */
+export function addSecondary (params) {
+  return executeCommand(addSecondaryEmail, [params.namedArgs.email]);
```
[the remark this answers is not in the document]
davlgd: imho, API errors should be parsed in sendAPI/request, let's discuss it in sync

src/parsers.js
```diff
@@ -69,6 +69,17 @@ export function futureDateOrDuration (dateString) {
   return duration;
 }
 
+// Could be:
```
aurrelhebert, line 72: Could or should?

hsablonniere:
## Review

I updated the commits to apply most of my feedbacks

### Usage:

* I don't think anyone will use this feature but why not
* I know primary is shorter than `mark-as-primary` but I think a verb/action would be better than a noun
* I don't think we need `open` but maybe we need to have a general discussion on how users interact with the CLI and the console
* I don't think we need the clear, most users don't even have a secondary email

### Code:

* executeCommand breaks the auto handling of errors with process exit 1
  * I removed it
  * I also put the server messages in the CLI so we can edit them if necessary
* `openBrowser` is a good idea, let's use it everywhere and add a helper for console URLs
* Let's try to avoid `(await somethingAsync()).someProp`, it's a bit harder to read
* I used the clever client (I know the endpoints are todo but it's OK)
* I merged the model in commands
* I added error handling with server error IDs
* I put most success message in the code (not from the API), same for some error messages
* I changed the way we handle errors in clear
* I changed the enveloppe emoji to ✉️

pdesoyres-cc (approved):
LGTM

PR #876, opened 2025-02-01
[PoC] Keycloak, Metabase, Otoroshi commands

davlgd:
⚠️ This is an experimental project to support Operators commands ⚠️


hsablonniere (approved):
Thanks @davlgd, this PR is really interesting regarding what I've been preparing for the new CLI as these 4 commands are very similar :+1: :clap: :muscle:

Here's what I changed on your PR:

* Simplify commands
  * Rename `keycloak ng enable` to `keycloak enable-ng`
  * Rename `keycloak ng disable` to `keycloak disable-ng`
  * Rename `otoroshi ng enable` to `otoroshi enable-ng`
  * Rename `otoroshi ng disable` to `otoroshi disable-ng`
  * It means no more `keycloak ng` and `otoroshi ng`
* Rename `src/lib/operators.js` to `src/lib/operator-commands.js`
  * Less files with the same name (like in models and clever-client)
  * A bit more precise to the fact that those are more or less the commands
* Rename `src/model/operators.js` to `src/model/operator.js`
  * Like we do for other models
* replaced `get(params)` with `await operatorPrint` in `ngEnable` and `ngDisable`
  * The lack of `await` or `return` was breaking the error handling chain
  * It's also cleaner not to have commands calling other commands directly
* Replace `openPage` with `openBrowser`
* Use `conf.GOTO_URL` here and there
* Adjust code for `ids-resolved/findAddonsByAddonProvider()`
  * Remove useless param
  * Rework the map/filter
* Adjust code for `operatorList`
  * Use lodash to capitalize
  * Remove 'addon-' from `providerName` for Matomo
    * prevents `🔎 Found 3 Addon-matomo operators`
    * prevents `🔎 No Addon-matomo found, create one with clever addon create addon-matomo command`
  * Use lodash for the groupBy (while we wait for a native replacement in Node.js 21)
* Fixes on `check version`
  * it was displaying `{ operator_id: 'CCC' }` when the user was giving a real ID
  * replacing tree like ascii with big dots for versions
  * Use the patched `confirm` prompt instead of a raw `select`
    * Avoid `User force closed the prompt with 0 null` on `Ctrl+C`
    * Exit with error on `no` like we do in other commands
* Add `'Content-Type': 'application/json'` to `versionUpdate` to get automatic JSON stringify
* Remove unused `listDeployed` and `setEnvVar`
* Remove unused `getOwnerIdFromOrgaIdOrName`
* This wording was a bit confusing for me:
  * `Manage operators managed services and their features such as Keycloak, Matomo, Metabase, Otoroshi`
  * => `Manage operators and their features such as Keycloak, Matomo, Metabase, Otoroshi`
* I introduced some `dedent` here and there, except when it would be harder to keep indentation

Notes :

* I only realized now that `findAddonsByNameOrId` partly does what `resolveRealId`
  * but `resolveRealId` does not handle names
  * and `findAddonsByNameOrId` does not use the IDs cache
  * I think we need to aim for a common implementation and stop relying on IDs starting with `keycloak_` or else
  * We'll do this in another iteration :wink:

PR #883, opened 2025-02-24
add clever tokens command to manage API tokens

hsablonniere:
### Create an API token

```
clever tokens create "[TOKEN NAME]"
```

You will be prompted for your password (and MFA code if necessary) and get this result:

```
✔ API token created successfully!

  - API token ID : [API_TOKEN_ID]
  - API token    : [API_TOKEN]
  - Expiration   : [EXPIRATION_DATE]

Export this token and use it to make authenticated requests to the Clever Cloud API through the Auth Bridge:

export CC_API_TOKEN=[API_TOKEN]
curl -H "Authorization: Bearer $CC_API_TOKEN" https://auth-bridge.clever-cloud.com/v2/self

Then, to revoke this token, run:
clever tokens revoke [API_TOKEN_ID]
```

### List API tokens

```
clever tokens
// or
clever tokens list
```

You will get this result:

```
┌─────────┬──────────────────┬─────────────┬─────────────────────┬────────────────────┬──────────────────────┐
│ (index) │ API token ID     │    Name     │ Creation IP address │ Creation date/time │ Expiration date/time │
├─────────┼──────────────────┼─────────────┼─────────────────────┼────────────────────┼──────────────────────┤
│    0    │ '[API_TOKEN_ID]' │ 'GitLab CI' │  'XXX.YY.ZZZ.123'   │ '2025-02-24 17:28' │  '2026-02-24 17:28'  │
└─────────┴──────────────────┴─────────────┴─────────────────────┴────────────────────┴──────────────────────┘
```

### Revoke token

```
clever tokens revoke [API_TOKEN_ID]
```

You will get this result:

```
✔ API token revoked successfully!
```

## Other details

This PR also adds https://www.npmjs.com/package/dedent as a dependency to ease the maintenance/indentation of multi-line messages.

PR #884, opened 2025-02-24
chore (tokens): add as experimental feature

davlgd:
- Add `tokens` as an experimental command
- Add documentation about `tokens` command
- Use `dedent` for experimental features info

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -993,7 +993,6 @@ async function run () {
     statusCommand,
     stopCommand,
     tcpRedirsCommands,
-    tokensCommands,
     versionCommand,
     webhooksCommand,
   ];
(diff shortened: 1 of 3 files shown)


hsablonniere:
@davlgd Thanks for this. I will add these in the main PR and squash the commits.

hsablonniere:
@davlgd FYI: `dedent` is a template tag function, it should be used without parens like this:

```
const myText = dedent`
  <div>foo</div>
`:
```

I'll update it.

hsablonniere:
I'm closing this PR as the commits are now in the main PR https://github.com/CleverCloud/clever-tools/pull/883

Thanks again @davlgd

PR #885, opened 2025-02-24
feat(tokens): add expiration option

davlgd:
Adds a `--expiration` or `-e` option to manage API token duration until expiration. Default value is `1y`, can be months (`M`), weeks (`w`) or days (`d`). Also `ms` or `s` but not documented as it's not very useful. Duration can't be greater than 1 year. 

```
clever create tokens myToken
clever create tokens myToken -e 6M
clever create tokens myToken --expiration 2y # Leads to an error
```

We could not use `duration-js` here, but it simplifies code.


hsablonniere:
@davlgd Thanks for this.

We already have two libs for duration in the project:

* `iso8601-duration`: to handle ISO durations (`PT10H` for 10 hours)
* `duration-js`: one for simple syntax (the one you used `10h` for 1 hours)

The problem is that `duration-js` thinks `m` means minutes and you used it for `months`.

I would prefer if we could have the same kind of behaviour as the other places where we accept such duration (logs and access logs with `--since` and `--until`).

My proposition would be to transform `--duration` into `--expiration` and accept:

* explicit dates
* simple durations (no month, no year)
* ISO durations (anything you need)

This way, it behaves like since/until and you can make sure a token will expire at a specific date.

We could also have 2 incompatible params: `--duration` and `--expiration`.

WDYT?

miton18:
I agree with @hsablonniere `expiration` is more explicit than `duration`
You can rely on date parser here https://github.com/CleverCloud/clever-tools/blob/master/src/parsers.js#L45

pdesoyres-cc:
I prefer two different parameters `--duration` and `--expiration`.

The command needs to check for incompatibility and throw error if both are specified.

* Every durations in the CLI should be treated the same way (same expected format).
* Every dates in the CLI should be treated the same way (same expected format).

src/commands/tokens.js
```diff
@@ -0,0 +1,139 @@
...
+  const createdToken = await createApiToken(tokenData).then(sendToAuthBridge);
+
+  switch (format) {
+    case 'json':
+      Logger.printJson(createdToken);
+      break;
+    case 'human':
+    default:
+      Logger.println(dedent`
+        ${colors.green('✔')} API token successfully created!
```
davlgd, line 73:
```suggestion
        ${colors.green('✔')} API token successfully created! Store it securely, you won't able to print it again.
```

src/experimental-features.js
```diff
@@ -12,6 +14,19 @@ export const EXPERIMENTAL_FEATURES = {
...
 
       Learn more about Materia KV: https://www.clever-cloud.com/developers/doc/addons/materia-kv/
-      `,
+    `,
+  },
+  tokens: {
+    status: 'beta',
+    description: 'Manage API tokens to query Clever Cloud API from https://auth-bridge.clever-cloud.com',
+    instructions: dedent`
+      Target any compatible add-on by its name or ID (with an org ID if needed) and send commands to it:
```
davlgd, line 23:
```suggestion
      Create, list or revoke API tokens from a single command:
```

src/experimental-features.js
```diff
@@ -12,6 +14,19 @@ export const EXPERIMENTAL_FEATURES = {
...
+
+          clever tokens create myTokenName
+          clever tokens list
```
davlgd, line 26:
```suggestion
          clever tokens list
          clever tokens list -F json
```

src/models/send-to-api.js
```diff
@@ -43,6 +57,9 @@ export function processError (error) {
   if (code === 'ECONNRESET') {
     throw new Error('The connection to the Clever Cloud API was closed abruptly, please try again.', { cause: error });
   }
+  if (error?.response?.status === 401) {
+    throw new Error('Not connected, use clever login or the env var to authenticate to the clever-token', { cause: error });
```
davlgd, line 61:
```suggestion
    throw new Error(`You're not logged in, use ${colors.red('clever login')} command to connect to your Clever Cloud account`, { cause: error });
```

davlgd (commented):
Some texts nitpicks

hsablonniere:
Changes were moved and integrated to https://github.com/CleverCloud/clever-tools/pull/883

PR #887, opened 2025-02-25
Feat/addon logs

miton18:


miton18:
up

StevenLeRoux:
@miton18 do you need the second item (feature flag) to be complete before merging it?

PR #895, opened 2025-03-06
fix: ignore org ID when using an app_id

### LINKED ISSUE #889: `clever link` with empty --org option and app full uid
If I tried to link my current folder to an app, but there is no `$CC_ORGA_ID` and my application ID is a full unique ID, I got these two messages.

```
$ clever --version
3.11.0
$ echo $CC_ORGA_ID
<empty>
$ clever link app_xxxx --org=$CC_ORGA_ID
[WARN] You've specified a unique application ID, organisation option will be ignored
[ERROR] Organisation not found
```

I'm in a GitLab Job, that's why I don't know if `$CC_ORGA_ID` is filled.
For now, I do a `test -n $CC_ORGA_ID` before executing the `clever link`

davlgd:
Try to link an app with and app ID and an empty or (in)valid org. The user is warned the org is ignored as before, but it works even if the org doesn't exist. 

fixes #889

### DIFF
--- a/src/commands/link.js
+++ b/src/commands/link.js
@@ -7,9 +7,11 @@ export async function link (params) {
 
   if (app.app_id != null && orgaIdOrName != null) {
     Logger.warn('You\'ve specified a unique application ID, organisation option will be ignored');
+    await Application.linkRepo(app, null, alias);
   }
-
-  await Application.linkRepo(app, orgaIdOrName, alias);
+  else {
+    await Application.linkRepo(app, orgaIdOrName, alias);
+  }
 
   Logger.println('Your application has been successfully linked!');
 }


chmuche:
:heart_hands:  Work like a charm :heart_hands: 

I use [git-aggregator](https://github.com/acsone/git-aggregator) to try this PR.


```yml
./clever-tools:
  remotes:
    clever-tools: https://github.com/CleverCloud/clever-tools.git
  merges:
    - clever-tools master
    - clever-tools davlgd-fix-link-org-app-id
```

davlgd:
Great, we have a planned released for later today, I'll check with team when we can review/merge this.

hsablonniere (approved):

PR #899, opened 2025-03-07
fix(deploy): fix app naming on deployment

### LINKED ISSUE #898: clever deploy don't support app with colon in name
I started to create applications and add-ons with colon in the name.
But it's impossible to use `clever deploy` on the applications.

I don't know if it's a bug of the `clever deploy` command or not. 
But it seems your remote git server uses internally the app name 😅 

### Sample

```shell
$ clever create "app:with:colon"
$ clever deploy
[ERROR] "app:with:colon" would be an invalid git reference. (Hint: a valid alternative would be "app-with-colon".)
```

### Workaround

Stop using colon `:` in your app name.

aurrelhebert:
closes #898

### DIFF
--- a/src/lib/slugify.js
+++ b/src/lib/slugify.js
@@ -1,0 +1,10 @@
+import slugifyRaw from 'slugify';
+
+/**
+ * Converts a string to a slug using strict mode
+ * @param {string} string - The string to be converted to a slug
+ * @returns {string} The slugified string
+ */
+export function slugify (string) {
+  return slugifyRaw(string, { strict: true });
+}
--- a/src/models/app_configuration.js
+++ b/src/models/app_configuration.js
@@ -2,7 +2,7 @@ import { promises as fs } from 'node:fs';
 import path from 'node:path';
 
 import _ from 'lodash';
-import slugify from 'slugify';
+import { slugify } from '../lib/slugify.js';
 
 import { Logger } from '../logger.js';
 import * as User from './user.js';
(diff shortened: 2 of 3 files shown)


aurrelhebert:
This PR suppress the `:` from the given app name as aliases to be able to use it as remote. The drawback of this fix is that an app with two aliases : `test:app` and `testapp`. An alternative could be to replace `:` by `%3A` for example or to use the AppId

hsablonniere (changes requested):
@aurrelhebert Thanks for handling this!

About the lib:

* I was a bit surprised that this wasn't already covered by the lib
* I had a quick look at [another lib](https://www.npmjs.com/package/@sindresorhus/slugify) I used in other projects but it has 2 sub dependencies so I'm not 100% convinced to switch
* I guess using the `remove` feature of `slufigy` seems OK but I'd rather use `strict: true`, it covers other special chars

About where we do this:

* I would prefer if we applied this directly in `app_configuration`
  * we could add the option here: `alias: alias || slugify(appData.name)`
* This way, aliases created from the app name would be the same as the created remote
  * Automatic aliases would be easier to use in a command (without the need to quote them)
  * The pb is you could still provide an alias with a `:` or another bad character

My proposition is:

* add the `strict:true` option in `app_configuration` so we get good automated aliases
* add the `strict:true` option in `git` so we handle special chars for people who chose their own alias
* rename the commit to reflect the change

> The drawback of this fix is that an app with two aliases : test:app and testapp could be mixing them. An alternative could be to replace : by %3A for example or to use the AppId

I don't think this is a problem that we should take care of.

hsablonniere (approved):
LGTM :+1: Thanks @aurrelhebert (and me :stuck_out_tongue:)

aurrelhebert:
> LGTM 👍 Thanks @aurrelhebert (and me 😛)

Thanks @hsablonniere it will be helpful 😛

PR #913, opened 2025-05-28
fix(unlink): remove default when unlinking default app

aurrelhebert:
When doing the QA of the #862, I discoverd that unlinking an application after setting it as default with `make-default` wasn't deleting the `default` key in `.clever.json`. This was then leading to an execution error execting user to manually clean its `.clever.json` file


hsablonniere:
@aurrelhebert You PR only fixes the `unlink` but the problem also happens with `delete`. Also, the file was read and written twice. I updated the commit to fix both those limitations.

hsablonniere (approved):
LGTM :+1: Thank you @aurrelhebert :wink:

PR #915, opened 2025-06-02
docs: update for 3.13 release

davlgd:
Depends on last 3.13 release adjustements

### DIFF
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -1,0 +1,82 @@
+# Clever Cloud Operators
+
+Operators allow you to deploy services as turnkey solutions on Clever Cloud. They provision resources, configure them and expose an API and tools to ease their management during their lifecycle. To add operators commands to Clever Tools, enable the `operators` feature:
+
+```
+clever features enable operators
+```
+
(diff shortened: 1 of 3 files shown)


docs/operators.md
```diff
@@ -0,0 +1,82 @@
+# Clever Cloud Operators
+
+Operators allow you to deploy services as turnkey solutions on Clever Cloud. They provision resources, configure them and expose an API and tools to ease their management during their lifecycle. To add operators commands to Clever Tools, enable the `operators` feature:
+
+```
+clever features enable operators
+```
+
+Then, you can use the `clever operators` command to manage Keycloak, Matomo, Metabase and Otoroshi instances on Clever Cloud. For example to list them deployed services:
```
hsablonniere, line 9: nitpick: can we remove the backtick, it seems like `clever operators` is a real command. I would go for this instead:

```md
Then, you can use the clever commands for operators to manage Keycloak, Matomo, Metabase and Otoroshi instances on Clever Cloud. For example to list them deployed services:
```

hsablonniere (approved):
A few nitpicks. Well done @davlgd :+1:

PR #916, opened 2025-06-02
Add JSON format to emails/ssh-keys list command

davlgd:
As we removed the `list` command, we should add `format` as private option to `emails` and `ssh-keys` to be able to `clever emails -F json` or `clever ssh-keys --format json`


hsablonniere (approved):
Good catch, LGTM :+1: Thanks @davlgd

PR #918, opened 2025-06-05
API Tokens: handle accounts with no password

florian-sanders-cc:
## What does this PR do?

- Fixes some nitpicks related to TypeChecking,
- Handles cases where the user has no password set and tries to create an API token:
  - show an error message similar to the one from the console,
  - provide a link to the console so users may add a password.

## How to review?

- Text & display need to be discussed with @davlgd & @hsablonniere but the code is ready to review
- Check the commits and trust the tests I've made with a passwordless account.


src/commands/tokens.js
```diff
@@ -13,10 +13,23 @@ import { promptPassword } from '../prompt-password.js';
...
   const { expiration, format } = params.options;
+  const user = await getCurrentUser();
+
+  if (!user.hasPassword) {
+    const apiTokenListHref = new URL('/users/me/api-tokens', conf.CONSOLE_URL).href;
+    Logger.println(dedent`
+      Your Clever Cloud account is linked via GitHub and has no password. Setting one is required to create API tokens.
+      Go to the following page in order to set your password:
+    `);
+    Logger.println(apiTokenListHref);
```
davlgd, line 29:
```suggestion
    Logger.println(dedent`
      Your Clever Cloud account is linked via GitHub and has no password. Setting one is required to create API tokens.
      To do so, go to the following URL: ${apiTokenListHref}
    `);
```

davlgd (approved):
Just a nitpick, thanks for this

hsablonniere (approved):
LGTM :+1: Well done :clap: 

I changed 2 things:

* renamed the first commit to remove the `commands/` from the scope
* transform the `Logger.println` to a `throw Error`

PR #926, opened 2025-06-25
[PoC] 3rd party zones

### LINKED ISSUE #345: Make clever-tools configurable for multizone purpose
We are implementing more and more private/satellite zones.
We need to easily configure clever-tools to use another api url.

How could we do that?

davlgd:
Fixes #345


BlackYoup:
Just tested, quick functional review:
- we should have arguments to set the values, interactive mode is great but for scripts the arguments will be nicer
- it asks for CONSOLE_TOKEN_URL, not sure what to put there
- Could we also provide oauth tokens (not consumers, tokens)? It would help when `clever login` won't be available.

davlgd:
> * it asks for CONSOLE_TOKEN_URL, not sure what to put there
Classic value is: 
```
CONSOLE_TOKEN_URL: 'https://console.clever-cloud.com/cli-oauth',
```

PR #946, opened 2025-08-04
feat(domain): get information in JSON format

davlgd:
This PR adds `--format` option to `clever domain` command:
- Without any change to `human` (default) format
- With details in JSON format thanks to `parseDomain()`
  - Favourite fqdn is `response.favourite` field
  - `vhosts` field contains each fqdn with its detailed structure, which can help in scripts


aurrelhebert (approved):
LGTM, command tests OK

PR #950, opened 2025-08-11
fix: show help context when no callback is set for command

### LINKED ISSUE #949: A command without a callback throws an error
A command not supplied with a callback, should show the help context.

Since the custom error-catching wrapper around `cliparse.command` expect the callback to be defined, we get the following error:

```
file:///home/youn/Code/clever-cloud/clever-tools/bin/clever.js:85
    const promise = commandFunction(...args);
                    ^

TypeError: commandFunction is not a function
    at Object.action (file:///home/youn/Code/clever-cloud/clever-tools/bin/clever.js:85:21)
    at cli.execute (/home/youn/Code/clever-cloud/clever-tools/node_modules/cliparse/src/cli.js:89:30)
    at cli.parse (/home/youn/Code/clever-cloud/clever-tools/node_modules/cliparse/src/cli.js:167:9)
    at run (file:///home/youn/Code/clever-cloud/clever-tools/bin/clever.js:1474:12)

Node.js v24.3.0
```

sehnryr:
Fixes #949 

The `cliparse.command` patch to catch errors expects the command callback to be defined, but it is not always wanted (e.g. we might not want the parent command of a group to do anything). To fix this, we need to skip the patch when the callback is `undefined`.

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -1331,10 +1331,6 @@ async function run () {
   const databaseCommand = cliparse.command('database', {
     description: 'List available databases',
     commands: [backupsCommand],
-  }, async () => {
-    console.info('This command is not available, you can try the following commands:');
-    console.info('clever database backups');
-    console.info('clever database backups download');
   });
 
   // Patch help command description
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -81,10 +81,16 @@ process.stdout.on('error', (error) => {
 const cliparseCommand = cliparse.command;
 
 cliparse.command = function (name, options, commandFunction) {
-  return cliparseCommand(name, options, (...args) => {
-    const promise = commandFunction(...args);
-    handleCommandPromise(promise);
-  });
+  let callback;
+
+  if (commandFunction !== undefined) {
+    callback = (...args) => {
+      const promise = commandFunction(...args);
+      handleCommandPromise(promise);
+    };
+  }
+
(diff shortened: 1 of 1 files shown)


davlgd:
In such a case it's not really a bug, a command need to be set with `cliparse.command` and a command name, description, arguments, options, etc. An empty command to print `help` is already available. Is there any specific use case you needed that for?

But in your example, to add a `test` command you should have something like: 

```
import * as test from '../src/commands/test.js';

...

const testCommand = cliparse.command('test', {
    description: 'Run tests',
    args: [args.testFiles],
  } test.run);
```

Add `testCommand` to `commands` array and an `export async function run () { }` in `src/commands/test.js` (it could use `params` as parameters to receive args/options).

davlgd (commented):
I'll let @hsablonniere tell if he's ok with this, otherwise LGTM it works for me and we could remove default list on some commands where it's not useful. Thanks for this.

Just a nitpick on how the change is done. Maybe we could be more concise with a 

```
    return cliparseCommand(name, options, commandFunction ?
      (...args) => {
        const promise = commandFunction(...args);
        handleCommandPromise(promise);
      } : undefined
    );
```

I'll let other CLI team members comment on this and decide.

hsablonniere:
Closing in favor of #976, it's a small rework of @sehnryr PR but on the Clever Cloud org (easier for rework and with the auto CI previews).

PR #951, opened 2025-08-19
fix: handle correctly add-ons option with a "-"

davlgd:
This PR fixes an issue with add-ons options parsing. Currently, when you pass an option with a `-` in it, it's not parsed correctly. We missed it mainly because it doesn't fail with only 1 option used. But with multiple options, the parsing problem leads to an error.

## Steps to reproduce

Create an Otoroshi add-on with multiple options: 

```
clever addon create otoroshi otoNgTest --option networkgroup=true --option routes-domain="routes.example.com" --option access-domain="example.com"
```

Options won't pass correctly.

## What the PR do

It adds a `(?:-\w+)*` to the options parsing regex to handle options with `-` correctly


hsablonniere (approved):
LGTM :+1:

PR #979, opened 2025-09-19
feat: add support for SSH/TCP access logs

davlgd:
This PR removes the `HTTP only` filter for access logs and detects wether an access log is `HTTP`, `SSH` or `TCP` if this information doesn't come from the API.


hsablonniere:
Superseded by #1139, which reworks this in three reviewable commits and keeps your work credited as co-author. Closing this one in favour of it.

PR #980, opened 2025-09-25
chore(logs): make json-stream format use one liner printer

miton18:


hsablonniere (approved):
LGTM :+1: 

Thank you @miton18 :clap:

PR #985, opened 2025-10-15
Add Network Groups add-ons support

davlgd:
Add-ons should now be linked to a Network Group with their Real ID. This PR updates this with clearer instructions in documentation and help. 

It also fix an issue with `get` command when a member is in multiple network groups, it returns the domain, which help to distinguish them (we'll certainly enhance duplicate management through API later).


KannarFr (changes requested):
Little changes required, otherwise LGTM.

PR #1000, opened 2025-11-19
fix: don't fail deployment if an application has no domain

### LINKED ISSUE #998: clever deploy exits with "couldn't find a domain name" with an application without domains
Running `clever deploy --follow` might exit with code 1 when an application has no domain names.

```
2025-11-18T17:42:10.828Z: Successfully deployed in 0 minutes and 41 seconds
[ERROR] Couldn't find a domain name
```

Not having domain names shouldn't quit the deploy command.

davlgd:
Fixes #998

### DIFF
--- a/src/models/log-v4.js
+++ b/src/models/log-v4.js
@@ -134,11 +134,17 @@ export async function watchDeploymentAndDisplayLogs(options) {
   }
 
   if (deploymentEnded.state === 'OK') {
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
LGTM, just a suggestion but I think I would keep what you did anyway 🤔

PR #1005, opened 2025-12-02
chore: update dependencies

hsablonniere:
## Summary
- Update patch-level dependencies
- Update minor-level dependencies  
- Override js-yaml to 4.1.1 to fix prototype pollution vulnerability (CVE-2023-2251)

## Test plan
- [x] `npm run lint` passes
- [x] CI tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "@rollup/plugin-json": "6.1.0",
     "@rollup/plugin-node-resolve": "16.0.3",
     "@types/mime-types": "3.0.1",
+    "@types/picomatch": "4.0.2",
     "@yao-pkg/pkg": "6.10.1",
     "eslint": "9.39.1",
     "globals": "16.5.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,44 +47,48 @@
...
+    "prettier": "3.7.3",
+    "prettier-plugin-organize-imports": "4.3.0",
+    "rollup": "4.53.3",
+    "simple-git": "3.30.0",
+    "tinyglobby": "0.2.15",
+    "typescript": "5.9.3"
+  },
+  "overrides": {
+    "js-yaml": "4.1.1"
   },
```
florian-sanders-cc: **question:** I couldn't see any difference with or without this override, seems like even without this override ESLint ends up pulling `js-yaml@4.1.1` 🤔 

Image

hsablonniere: You're absolutely right. I'll remove it.

PR #1020, opened 2026-01-16
fix(link): error when linking apps with duplicate alias

### LINKED ISSUE #222: When you link a local git repo to 2 different clever apps, you get the same alias in .clever.json

hsablonniere:
## Summary
- When linking a second app that generates the same alias as an existing linked app, throw an error asking the user to specify a different alias with `--alias`
- Previously, duplicate aliases were silently created in `.clever.json`, causing confusion

Fixes #222

## Test plan
- [x] Link an app named "my-app" → should work, alias = "my-app"
- [x] Link another app also named "my-app" → should error with message to use `--alias`
- [x] Link that second app with `--alias my-app-2` → should work

### DIFF
--- a/src/models/app_configuration.js
+++ b/src/models/app_configuration.js
@@ -30,21 +30,30 @@ export async function loadApplicationConf(ignoreParentConfig = false, pathToFold
 export async function addLinkedApplication(appData, alias, ignoreParentConfig) {
   const currentConfig = await loadApplicationConf(ignoreParentConfig);
 
+  const generatedAlias = alias || slugify(appData.name);
+
+  const existingApp = currentConfig.apps.find((app) => app.app_id === appData.id);
+  if (existingApp != null) {
+    throw new Error(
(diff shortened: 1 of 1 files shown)


src/models/app_configuration.js
```diff
@@ -30,22 +30,32 @@ export async function loadApplicationConf(ignoreParentConfig = false, pathToFold
 export async function addLinkedApplication(appData, alias, ignoreParentConfig) {
   const currentConfig = await loadApplicationConf(ignoreParentConfig);
 
+  const generatedAlias = alias || slugify(appData.name);
+
+  const isPresent = currentConfig.apps.find((app) => app.app_id === appData.id) != null;
+  if (isPresent) {
+    const existingApp = currentConfig.apps.find((app) => app.app_id === appData.id);
```
florian-sanders-cc: **nitpick/suggestion:** 
```suggestion
  const existingApp = currentConfig.apps.find((app) => app.app_id === appData.id);
  if (existingApp != null) {
```

PR #1029, opened 2026-01-23
feat(config-provider): add commands to manage configuration providers

### LINKED ISSUE #1027: Add env var to Config-Provider with cli
The command `clever env set` is only available for application , so doesn't work for the Config Provider addon

hsablonniere:
Closes #1027

## Context

Configuration providers are a special type of add-on on Clever Cloud that allow users to store and manage environment variables independently from applications. These variables can then be linked to one or more applications, making it easy to share configuration across multiple services.

Unlike regular add-ons (databases, caches, etc.), configuration providers don't run any service - they simply hold key-value pairs that get injected as environment variables into linked applications.

## Proposal

This PR adds a new `clever config-provider` command group to manage configuration providers from the CLI.

### Commands

| Command | Description |
|---------|-------------|
| `clever config-provider` | Display help (doc-only) |
| `clever config-provider list` | List all configuration providers |
| `clever config-provider get <id>` | List environment variables |
| `clever config-provider set <id> <name> <value>` | Add or update a variable |
| `clever config-provider rm <id> <name>` | Remove a variable |
| `clever config-provider import <id>` | Import variables from stdin |
| `clever config-provider open <id>` | Open in Clever Cloud Console |

### Design decisions

**Command structure**: We use a flat structure (`clever config-provider get/set/rm/...`) for simplicity and ease of use, similar to how `clever env` works for applications.

**Identifier support**: The `<id>` argument accepts three formats (similar to operators):
- Add-on ID: `addon_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
- Real ID (config provider ID): `config_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
- Add-on name (if unambiguous)

**Consistency**: The commands follow the same patterns as `clever env` and `clever addon env`:
- Same output formats (human, json, shell)
- Same import behavior (replaces all variables)
- Same error handling

Also, the list display follows the same format as the operators.

## QA Testing

```bash
# List all configuration providers
clever config-provider list
clever config-provider list --format json

# Test with different ID formats
clever config-provider get <addon-id>
clever config-provider get <config_xxx-real-id>
clever config-provider get <addon-name>

# Set and remove variables
clever config-provider set <id> MY_VAR "my value"
clever config-provider get <id>  # verify it's set
clever config-provider rm <id> MY_VAR
clever config-provider get <id>  # verify it's removed

# Import from stdin (replaces all variables)
echo 'FOO="bar"' | clever config-provider import <id>
echo '[{"name": "BAZ", "value": "qux"}]' | clever config-provider import <id> --format json

# Open in browser
clever config-provider open <id>
```

### Expected behavior
- All three ID formats (addon_id, config_id, name) should work
- Import replaces all existing variables (with warning in help text)
- Error if the add-on is not a configuration provider
- Error if name is ambiguous (multiple matches)

### DIFF
--- a/src/commands/config-provider/config-provider.command.js
+++ b/src/commands/config-provider/config-provider.command.js
@@ -1,0 +1,10 @@
+import { defineCommand } from '../../lib/define-command.js';
+
+export const configProviderCommand = defineCommand({
+  description: 'Manage configuration providers',
+  since: 'unreleased',
+  options: {},
+  args: [],
+  // Parent command - no handler, only contains subcommands
+  handler: null,
+});
--- a/src/models/config-provider.js
+++ b/src/models/config-provider.js
@@ -1,0 +1,34 @@
+import { get as getAddon } from '@clevercloud/client/esm/api/v2/addon.js';
+import { findAddonsByNameOrId } from './ids-resolver.js';
+import { sendToApi } from './send-to-api.js';
+
+/**
+ * Resolve a config provider ID from a name, ID or real ID
+ * @param {string} addonIdOrRealIdOrName The add-on ID (addon_xxx), real ID (config_xxx) or name
+ * @returns {Promise<{ ownerId: string, realId: string, addonId: string }>} The owner ID, real ID and addon ID
+ * @throws {Error} If the add-on is not found
+ * @throws {Error} If multiple add-ons are found with the same name
+ * @throws {Error} If the add-on is not a configuration provider
+ */
+export async function resolveConfigProviderId(addonIdOrRealIdOrName) {
+  const candidates = await findAddonsByNameOrId(addonIdOrRealIdOrName);
+
+  if (candidates.length === 0) {
+    throw new Error(`Config provider not found: ${addonIdOrRealIdOrName}`);
+  }
+
+  if (candidates.length > 1) {
+    throw new Error(`Ambiguous config provider name '${addonIdOrRealIdOrName}', please use the ID`);
+  }
+
+  const { ownerId, addonId, realId } = candidates[0];
+
+  // Verify that the addon is a config provider
+  const addon = await getAddon({ id: ownerId, addonId }).then(sendToApi);
+
+  if (addon.provider.id !== 'config-provider') {
+    throw new Error(`The add-on '${addonIdOrRealIdOrName}' is not a configuration provider`);
+  }
+
+  return { ownerId, realId, addonId };
+}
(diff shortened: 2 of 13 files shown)


src/models/config-provider.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Resolve a config provider ID from a name, ID or real ID
+ * @param {string} addonIdOrRealIdOrName The add-on ID (addon_xxx), real ID (config_xxx) or name
+ * @returns {Promise<{ ownerId: string, realId: string, addonId: string }>} The owner ID, real ID and addon ID
+ * @throws {Error} If the add-on is not found
+ * @throws {Error} If multiple add-ons are found with the same name
+ * @throws {Error} If the add-on is not a configuration provider
+ */
+export async function resolveConfigProviderId(addonIdOrRealIdOrName) {
+  const candidates = await findAddonsByNameOrId(addonIdOrRealIdOrName);
```
florian-sanders-cc: **note:** the JSDoc for `findAddonsByNameOrId` is wrong and makes LSP report an error (doc says it returns an object + no promise).
I guess we'll tackle this when we move to TypeScript anyway so it's fine?

src/commands/config-provider/config-provider.command.js
```diff
@@ -0,0 +1,45 @@
+import { defineCommand } from '../../lib/define-command.js';
+import { styleText } from '../../lib/style-text.js';
+import { Logger } from '../../logger.js';
+import { findAddonsByAddonProvider } from '../../models/ids-resolver.js';
+import { humanJsonOutputFormatOption } from '../global.options.js';
+
+export const configProviderCommand = defineCommand({
+  description: 'List configuration providers',
+  since: '4.6.0',
```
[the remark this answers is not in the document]
hsablonniere: I created this issue. We'll figure it out there, in the meantime, I'll keep this PR as is.

https://github.com/CleverCloud/clever-tools/issues/1039

src/models/config-provider.js
hsablonniere: > I guess we'll tackle this when we move to TypeScript anyway so it's fine?
Yes, clearly this will be done anyway in a later TypeScript step.

Do you think I should update the JSDoc to `Promise<Object>` in this PR in its own commit?

florian-sanders-cc: Unless you want to tackle it right now, I think it's fine to fix this issue later on when we move to TypeScript anyway :+1:

PR #1035, opened 2026-01-29
refactor(logger): remove lodash, add types

hsablonniere:
## Context

The logger module is imported throughout the codebase. Since it wasn't typechecked, **it was blocking the addition of many files to `tsconfig.json`**.

Other issues:
- **Lodash dependency**: used `_.map()`, `_.get()`, `_.repeat()`, `_.noop()`, `_.fromPairs()` for trivial operations that can be done with native JS
- **Dynamically generated code**: the `debug/info/warn/error` methods were generated via a lodash chain, making the code hard to read and impossible to type properly

## Changes

- **Add JSDoc types**: all functions are now typed with proper annotations
- **Add `logger.types.d.ts`**: defines the `ApiError` interface for API errors
- **Remove lodash**: replaced with native JS (`Object.entries()`, `' '.repeat()`, `instanceof`, etc.)
- **Explicit code**: each Logger method is now defined explicitly instead of being dynamically generated

## Later

This refactor is intentionally **iso-functional**. The goal is to clean up without going down an endless refactoring rabbit hole.

Possible improvements for later:
- Rename `Logger` to `logger` (lowercase convention for singletons) — not done because it would generate a large diff across all importing files, and we have several PRs in progress
- Use a more powerful logger shared across Clever Cloud JS projects

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,85 +1,139 @@
-import _ from 'lodash';
 import { format } from 'node:util';
 import { styleText } from './lib/style-text.js';
 
-function getPrefix(severity) {
+/**
+ * @typedef {import('./logger.types.js').ApiError} ApiError
+ */
+
+const IS_QUIET = Boolean(process.env.CLEVER_QUIET);
+const IS_VERBOSE = Boolean(process.env.CLEVER_VERBOSE);
+
+export const Logger = {
+  /**
+   * @param {string} message
+   */
+  debug(message) {
+    consoleLog('debug', message);
+  },
+
+  /**
+   * @param {string} message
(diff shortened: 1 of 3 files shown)


src/logger.js
```diff
@@ -1,85 +1,139 @@
...
+  },
 
-function consoleErrorWithoutColor(line) {
-  process.stderr.write(format(line) + '\n');
-}
+  /**
+   * @param {string} message
+   */
+  info(message) {
+    return consoleLog('info', message);
```
pdesoyres-cc: nitpick: do we need `return`?

PR #1043, opened 2026-02-04
fix(addon): fix Matomo add-on creation

### LINKED ISSUE #1042: Bad request during Matomo creation
When I create a Matomo, it fails after creation.

Problem is that CLI tries to query `https://api.clever-cloud.com/v4/addon-providers/addon-addon-matomo/addons/matomo_id`

If you replace `addon-addon` with `addon` it's ok.

hsablonniere:
Closes #1042

## Context

Creating a Matomo addon failed with a 404 error because the operator API path was constructed incorrectly. The v4 operator APIs use paths like `/v4/addon-providers/addon-${provider}/addons/...`, which adds an `addon-` prefix. For
providers like `addon-matomo`, passing the full provider name resulted in `addon-addon-matomo`.

This inconsistency in provider naming only affects Matomo among the operators (keycloak, metabase, otoroshi don't have the `addon-` prefix in their keys).

## Solution

Replace the `isOperator` boolean with an explicit `operatorProvider` string that specifies the exact provider name to use in API calls. This allows each provider to declare the correct string needed for the operator API path, avoiding the duplicate prefix issue.

- `addon-matomo` → `operatorProvider: 'matomo'`
- `keycloak` → `operatorProvider: 'keycloak'` (unchanged behavior)
- Non-operator addons (`kv`, `addon-pulsar`) simply omit the property

### DIFF
--- a/src/commands/addon/addon.create.command.js
+++ b/src/commands/addon/addon.create.command.js
@@ -20,12 +20,11 @@ import { addonNameArg, addonProviderArg } from './addon.args.js';
 const ADDON_PROVIDERS = {
   keycloak: {
     name: 'Keycloak',
-    isOperator: true,
+    operatorProvider: 'keycloak',
     postCreateInstructions: `Learn more about Keycloak on Clever Cloud: ${config.DOC_URL}/addons/keycloak/`,
   },
   kv: {
     name: 'Materia KV',
-    isOperator: false,
     status: 'beta',
     postCreateInstructions: (addonId) => dedent`
       ${styleText('yellow', "You can easily use Materia KV with 'redis-cli', with such commands:")}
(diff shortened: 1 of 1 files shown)


florian-sanders-cc (approved):
LGTM, some suggestions to match our conventions & make it slightly more robust but it could be merged as it is (tested locally as well just in case 😉)

florian-sanders-cc (changes requested):
Ah! Forgot to check the commit message: Could we reword it so that users (not contributors to this project) understand the issue that it fixes for them? (Again, this is an issue with using our conventional commits for changelog, consumers don't expect the same info as reviewers / contributors)

Something that would mention that we've fixed Matomo addon creation :thinking:

PR #1051, opened 2026-02-18
feat(ssh): add interactive instance selection and --command option

davlgd:
## Context

The SSH proxy's built-in instance selection relies on a raw number input, which is a poor UX. Additionally, this mechanism is incompatible with non-interactive command execution (piped stdin). Both issues are solved by moving instance selection client-side via the API.

## Proposal

Three incremental changes:

1. **Refactor SSH params construction** — build the params array progressively to prepare for conditional flags (`-t`, `-i`, target)

2. **Client-side instance selection** — query running instances via the API instead of relying on the SSH proxy's number-based prompt:
   - Single instance → connect directly
   - Multiple instances in TTY → interactive select prompt
   - Multiple instances in non-TTY → error out
   - No instances → error out

3. **`-c, --command` option** — execute a single command on the remote instance and exit. Since the gateway uses the remote command slot for routing (appId), the command is written to stdin. A UUID marker delimits real output from gateway/login noise. PTY allocation is skipped in this mode to keep stdout clean.

## How to review

All changes are in `src/commands/ssh/ssh.command.js`:
- Start with the refactor (param building logic)
- Then the instance selection block (API call + prompt)
- Then the command execution branch (marker-based output filtering)

## How to test

- `clever ssh` with a single running instance → connects directly
- `clever ssh` with multiple instances → shows a selection prompt
- `clever ssh -c 'echo $APP_ID'` → prints the command output and exits
- `clever ssh -c 'echo $APP_ID'` with multiple instances → prompts then executes
- Pipe mode: `clever ssh -c 'echo $APP_ID' | cat` → works without TTY

### DIFF
--- a/skills/clever-tools/references/full-documentation.md
+++ b/skills/clever-tools/references/full-documentation.md
@@ -3065,6 +3065,7 @@ clever ssh [options]
 ```
 -a, --alias <alias>                    Short name for the application
     --app <app-id|app-name>            Application to manage by its ID (or name, if unambiguous)
+-c, --command <command>                Execute a command on the remote instance and exit
 -i, --identity-file <identity-file>    SSH identity file
 ```
 
--- a/src/commands/ssh/ssh.command.js
+++ b/src/commands/ssh/ssh.command.js
@@ -1,5 +1,6 @@
 import { getAllInstances } from '@clevercloud/client/esm/api/v2/application.js';
 import { spawn } from 'node:child_process';
+import { randomUUID } from 'node:crypto';
 import { z } from 'zod';
 import { config } from '../../config/config.js';
 import { defineCommand } from '../../lib/define-command.js';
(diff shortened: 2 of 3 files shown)


pdesoyres-cc (changes requested):
Well done. Nice improvement.

I left some minor issues/ideas.

I don't know why and how we could fix that, but when I use the `-c` like this:

```
➜ cleverr ssh --app app_xxx -c ls
✔ Select an instance: Crawling ursaring - Instance 1 - UP (xxx)
app_xxx
application_id
applicationrc
cargo-home
cargostarter-1.0.jar
containers
envlist
extra-libs
owner_id
zabbix-scripts
```

but when I enter in interactive mode inside the VM, the `ls` command shows a one line output:
```
➜ cleverr ssh --app app_xxx
✔ Select an instance: Crawling ursaring - Instance 1 - UP (xxx)
Opening an ssh shell.
Warning: Permanently added '[xx.xx.xx.xx]:27956' (ED25519) to the list of known hosts.
Last login: Thu Mar  5 17:06:36 UTC 2026
bas@xxx ~ $ ls
app_xxx  application_id  applicationrc  cargo-home  cargostarter-1.0.jar  containers  envlist  extra-libs  owner_id  zabbix-scripts
```

pdesoyres-cc (approved):
LGTM

PR #1073, opened 2026-03-16
feat(drain): add test-command subcommand

KannarFr:
## Summary

- Add `getDrainTestCommand` API client method for the `GET .../drains/{drainId}/test-command` endpoint
- Add `clever drain test-command <drain-id>` subcommand that returns a ready-to-execute shell command (curl/netcat) to manually test a drain recipient

## Test plan

- [ ] `clever drain test-command <drain-id> --app <app>` returns a valid shell command for HTTP drains (curl)
- [ ] `clever drain test-command <drain-id> --app <app>` returns a valid shell command for syslog drains (netcat)
- [ ] `clever drain test-command --help` displays correct usage
- [ ] Other drain commands (`create`, `get`, `disable`, `enable`, `remove`) still work with `--app`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/clever-tools/references/full-documentation.md
+++ b/skills/clever-tools/references/full-documentation.md
@@ -1170,6 +1170,28 @@ drain-id                       Drain ID
     --app <app-id|app-name>    Application to manage by its ID (or name, if unambiguous)
 ```
 
+### drain test-command
+
+**Description:** Get a ready-to-execute shell command to test a drain recipient
+
+**Since:** 4.8.0
+
+**Usage**
+```
+clever drain test-command <drain-id> [options]
+```
+
+**Arguments**
+```
+drain-id                       Drain ID
+```
+
+**Options**
+```
+-a, --alias <alias>            Short name for the application
+    --app <app-id|app-name>    Application to manage by its ID (or name, if unambiguous)
+```
+
 ## emails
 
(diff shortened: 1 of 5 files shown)


KannarFr:
> > Revert --resource option on drain commands, use --app only for now
> 
> Is there a specific reason for this?

No, it was bad branch management, removing it from overview.

hsablonniere:
@KannarFr Can you give more context?

From what I understood, we output a shell command that users can run to call their own app to mimick what we do when we send logs to drains.

1. Why don't we do the call on the backend side ourselves?
2. Won't we have differences between what we really do (headers, IPs...)?

I expected a similar UX as when I setup a hook on GitHub for example.

image

Naming nitpick: I expected `test-command` to test something (and do a call on the backend side), no to output (get) et command.

KannarFr:
Yes I agree on your statements, will do.

KannarFr:
Closing this — the test-command logic should live on the API side rather than being exposed as a CLI subcommand. The backend endpoint returns a ready-to-execute shell command, but the actual testing/validation of a drain recipient is better handled server-side so it benefits all API clients (console, CLI, third-party) consistently, rather than being reimplemented per client.

PR #1074, opened 2026-03-18
fix(ng): adapt to @clevercloud/client v12 API change for peer config

pdesoyres-cc:
## Summary
- Upgrade `@clevercloud/client` from v11.3.0 to v12.0.0
- Adapt NG peer config handling to the new API response format (plain text instead of base64-encoded JSON)
- Remove the --format option (no JSON format anymore)

## Test plan
- [ ] Run `clever ng get-config` and verify the WireGuard configuration is displayed as plain text

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@babel/traverse": "7.28.5",
-        "@clevercloud/client": "11.3.0",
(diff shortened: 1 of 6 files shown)


pdesoyres-cc:
@davlgd. Considering your review, I removed the support of json format for this command.

hsablonniere (approved):
Tested locally with up to date client and it works.

LGTM 👍

PR #1077, opened 2026-03-27
feat: add oauth-consumers commands

### LINKED ISSUE #1079: config-provider list: align output style with operatorList (emoji + bullet points)
## Context / Problem

The `config-provider list` command outputs a list grouped by organisation, similar to operator commands (`otoroshi list`, `keycloak list`, `matomo list`, `metabase list`), but doesn't follow the same formatting style.

**`clever otoroshi`** (via `operatorList` in `src/lib/operator-commands.js`):
```
🔎 Found 2 Otoroshi operators:

• orga_xxx (ACME FOO)
  • fs-test-operator-otoroshi (otoroshi_xxx)

• orga_yyy (ACME BAR)
  • test-addon-otoroshi (otoroshi_yyy)
```

**`clever config-provider list`** (`src/commands/config-provider/config-provider.list.command.js`):
```
Found 1 configuration provider:

orga_yyy (ACME BAR)
  test-addon-config (config_xxx)
```

Missing:
- `🔎` emoji prefix on the header line
- `•` bullet points on org lines and item lines

## TODO

- [ ] Add `🔎` emoji prefix to the header line (line 32) and the empty case (add a line similar to operatorList)
- [ ] Add `•` before each org group header (line 36)
- [ ] Add `•` before each provider item (line 38)
- [ ] Consider extracting a generic helper for displaying grouped-by-org lists (the pattern is duplicated between `operatorList` in `src/lib/operator-commands.js` and `config-provider list` — a shared helper would ensure consistent formatting and reduce duplication for future similar commands)

davlgd:
## Context
Add CLI commands to manage OAuth consumers directly from `clever-tools`. This enables users to create, list, get, update, delete, and open OAuth consumers — both for personal accounts and organisations.

This is a rework of the original implementation by @davlgd with a focus on consistency with the rest of the codebase.

## Proposal
The PR is structured as preparatory refactorings followed by the feature itself:

### Refactoring (6 commits)
- Remove redundant properties (`options: {}`, `args: []`, `handler: null`) from parent namespace commands
- Normalize `openBrowser` messages across all commands (centralize 🌐 prefix, consistent wording)
- Rename `confirmAddonDeletionOption` → `skipConfirmationOption` for broader reuse
- Consolidate owner ID resolution into `getOwnerIdFromOrgIdOrName` (from `ids-resolver`)
- Extract `printItemsByOwner` utility (shared by config-provider list and operator list)
- Add `promptCheckbox`, `promptTextOption`, and `unwrapSchema` utilities

### Feature (1 commit)
- New `oauth-consumers` command group with 6 subcommands: `create`, `delete`, `get`, `list`, `open`, `update`
- Consumer resolution by key or name (with disambiguation prompt if name is ambiguous)
- Interactive rights selection via checkbox prompt for `create` and `update`
- Support for `--format json` output and `--org` targeting
- Model layer with full CRUD via Clever Cloud API (`/v2/oauth/consumers`)

## How to review
1. Start with the refactoring commits (oldest first) — each is self-contained
2. Then review the feature commit, starting with `src/models/oauth-consumer.js` (API layer) and `src/commands/oauth-consumers/` (command definitions)

### DIFF
--- a/skills/clever-tools/references/full-documentation.md
+++ b/skills/clever-tools/references/full-documentation.md
@@ -348,7 +348,7 @@ addon-id|addon-name                     Add-on ID (or name, if unambiguous)
 **Options**
 ```
 -o, --org, --owner <org-id|org-name>    Organisation to target by its ID (or name, if unambiguous)
--y, --yes                               Skip confirmation and delete the add-on directly
+-y, --yes                               Skip confirmation and proceed with deletion directly
 ```
 
 ### addon env
@@ -1530,7 +1530,7 @@ cluster-id|cluster-name                 Kubernetes cluster ID or name
 **Options**
 ```
 -o, --org, --owner <org-id|org-name>    Organisation to target by its ID (or name, if unambiguous)
--y, --yes                               Skip confirmation and delete the add-on directly
+-y, --yes                               Skip confirmation and proceed with deletion directly
 ```
 
 ### k8s get
--- a/skills/clever-tools/references/full-documentation.md
+++ b/skills/clever-tools/references/full-documentation.md
@@ -2508,6 +2508,146 @@ notification-id                         Notification ID
 -o, --org, --owner <org-id|org-name>    Organisation to target by its ID (or name, if unambiguous)
 ```
 
+## oauth-consumers
+
+**Description:** Manage OAuth consumers used with a Clever Cloud login
+
+**Since:** unreleased
+
+**Usage**
+```
+clever oauth-consumers
+```
+
+### oauth-consumers create
+
+**Description:** Create an OAuth consumer
+
+**Since:** unreleased
+
+**Usage**
+```
+clever oauth-consumers create <name> [options]
+```
+
+**Arguments**
+```
+name                                    Consumer name
+```
+
+**Options**
+```
+    --base-url <url>                    OAuth callback base URL
+-d, --description <description>         Consumer description
+-F, --format <format>                   Output format (human, json) (default: human)
+-o, --org, --owner <org-id|org-name>    Organisation to target by its ID (or name, if unambiguous)
+    --picture <url>                     Application logo URL
+    --rights <rights>                   Comma-separated list of rights (access-organisations, access-organisations-bills, access-organisations-consumption-statistics, access-organisations-credit-count, access-personal-information, manage-organisations, manage-organisations-applications, manage-organisations-members, manage-organisations-services, manage-personal-information, manage-ssh-keys, all)
+    --url <url>                         Application home URL
+```
+
+### oauth-consumers delete
+
+**Description:** Delete an OAuth consumer
+
+**Since:** unreleased
+
+**Usage**
+```
+clever oauth-consumers delete <consumer-key|consumer-name> [options]
+```
+
+**Arguments**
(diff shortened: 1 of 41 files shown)


hsablonniere (changes requested):
Thanks @davlgd, I did a UX review, I'll do a code review after, once we agree (or not 😛) on the following points.

## list
  
In commands for operators or specific add-ons like config-provider, the list displays all items across all organisations. I think we should do the same. It can be done with the summary. This would also fix the weird indentation for the items with dots.

## create (and update)

I wonder if the name should be an option instead of an arg (like for update) and it could even be part of the interactif stuffs. WDYT?

At first, I was like "why are all these options optional?" but now I get it, not specified means interactive mode. It will be nice to have something more declarative for this in our future CLI framework :wink:

- The `--with-secret` is nice
- I like the update with interactive mode

I was wondering if we could have `access` and `manage` as jokers but `all` is clearly enough for such a low usage feature.

## get/delete/update

I think the org should be removed, we can rely on the key (and use the summary, like I suggested for the list).

## other

I noticed the `config-provider` has some display differences with how operators are listed. I created an [issue](https://github.com/CleverCloud/clever-tools/issues/1079). Maybe this PR can address this in a dedicated commit and then reuse the helper to format the list for OAuth consumers.

davlgd:
Thanks for the thorough UX review! I've pushed new commits to fix issues, some related to your comment. 

Here are my thoughts on each point:                                                                                 
                                                               
> In commands for operators or specific add-ons like config-provider, the list displays all items across all organisations. I think we should do the same. It can be done with the summary. This would also fix the weird indentation for the items with dots.
                                                                                                                                         
Agreed, I've aligned with the config-provider/operators pattern: list across all orgs, grouped by owner with the summary format.

> I wonder if the name should be an option instead of an arg (like for update) and it could even be part of the interactif stuffs. WDYT?

I'd prefer keeping `name` as a positional arg. Every create command in clever-tools (`addon`, `drain`, `ng`, `k8s`, `tokens`…) uses the primary identifier as a positional arg — I'd rather stay consistent with that established pattern.

The difference with update is intentional: in create you **name**, in update you optionally **rename**, hence arg vs option.
                               
> I think the org should be removed, we can rely on the key (and use the summary, like I suggested for the list).                                      
   
Makes sense, the consumer key is unique so `--org` is redundant for get/delete/update. I've resolved the owner internally from the summary instead.     
                                                               
> I noticed the config-provider has some display differences with how operators are listed. I created an issue. Maybe this PR can address this in a dedicated commit and then reuse the helper to format the list for OAuth consumers.
                                                                                                 
Good idea, I've added a dedicated commit to extract a shared list formatting helper and reuse it for oauth-consumers. That way we address #1079 and keep the list display consistent.

hsablonniere:
Thanks for your feedbacks @davlgd 

> I'd prefer keeping name as a positional arg...

Agreed, I'll do the code review soon.

hsablonniere:
Hey @davlgd, I reworked the implementation to better align with the rest of the codebase (consistency patterns, shared utilities, etc.). I also updated the PR description to reflect the current state of the code.

hsablonniere (approved):
LGTM 👍

PR #1082, opened 2026-04-21
feat(logger): persistent file logging and structured API errors

pdesoyres-cc:
## Context

The previous `Logger` only wrote to stdout, so output vanished as soon as the process exited — bug
reports were impossible to triage after the fact. API errors went through an ad-hoc `processError`
that rendered a `fields` map, hid the raw response body, and made support tickets painful. A long
tail of direct `console.*` calls in commands meant `CLEVER_QUIET` / `CLEVER_VERBOSE` applied
unevenly across the CLI.

## Changes

- Adopt `@clevercloud/scribe` and persist every log level to a rotating file at OS-conventional
  paths: `$XDG_STATE_HOME/clever-cloud/` (Linux), `~/Library/Logs/clever-cloud/` (macOS),
  `%LOCALAPPDATA%\clever-cloud\Logs\` (Windows). Rotation kicks in around 250k, capped at 50 files.
- Introduce `ApiError extends Error` exposing `code`, `status`, raw `body`, `url`, `headers`. The
  legacy `processError` `fields` rendering is gone — `ApiError.body` carries the raw response.
- Unify the pretty-print path: a single `prettyLog` handles every severity, including stack traces
  under `CLEVER_VERBOSE=1` and styled `[ERROR]` / `[WARN]` / `[INFO]` / `[DEBUG]` prefixes.
- Migrate the remaining direct `console.log` / `console.table` / `console.error` / stderr writes
  through `Logger.println` / `printTable` / `printErrorLine` / `warn`, and split user-facing
  warnings from the internal `warn` channel.
- Route every exit through a new `src/lib/exit.js` helper that awaits scribe's `shutdown()` before
  calling `process.exit`, so the file transport's async buffer is flushed on success, failure,
  EPIPE, SIGINT/SIGTERM, prompt cancellation, and ssh/curl/config error paths.
- Add ESLint rules scoped to `bin/*.js` and `src/**/*.js`: `no-console` to lock in the Logger
  migration, and a `no-restricted-properties` rule forbidding `process.exit` outside the new
  exit helper.

### Implementation notes

The pretty-print path was rewritten rather than patched: the old code had divergent branches for
`error` (red prefix, stack trace, stderr) and the other severities (no styling, stdout, gated on
`IS_VERBOSE`). Routing every severity through `prettyLog` makes the `IS_QUIET` / `IS_VERBOSE` and
stderr-vs-stdout decisions explicit instead of scattered.

Cliparse's own parse-time `process.exit(1)` calls are deliberately left untouched — they fire
before any meaningful logging happens, so the lost file lines aren't worth a global `process.exit`
shim. The ESLint rule's `ignores` entry exempts `src/lib/exit.js` itself, which is the one
sanctioned caller of `process.exit`.

## How to review

1. Start with `src/logger.js` — `prettyLog` and `getLogFilePath` are the load-bearing changes.
2. Read `src/lib/api-error.js` plus `toApiError` / `parseError*` in `src/models/send-to-api.js`
   to see how API failures now flow to the user.
3. Read `src/lib/exit.js` and the call sites in `bin/clever.js`, `src/lib/cliparse-patched.js`,
   `src/lib/prompts.js`, and the ssh/curl/config commands to confirm every exit path flushes.
4. Run any `clever` command and confirm a log file appears under the OS-specific path above; force
   a few rotations to check the 250k / 50-files cap.
5. `CLEVER_QUIET=1` suppresses stdout but the file is still written; `CLEVER_VERBOSE=1` surfaces
   `[DEBUG]` / `[INFO]` entries on stdout and prints full stack traces for thrown `Error`s.
6. Trigger a 4xx/5xx API call and confirm the user sees `<message> [<code>]` and that
   `ApiError.body` carries the raw response payload.
7. Send SIGINT mid-command and verify the last log lines actually land in the file (i.e. the
   buffer was flushed before exit).

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -16,6 +16,7 @@ import {
 } from '../src/commands/global.options.js';
 import { EXPERIMENTAL_FEATURES, getFeatures } from '../src/config/features.js';
 import { cliparse } from '../src/lib/cliparse-patched.js';
+import { exit } from '../src/lib/exit.js';
 import { styleText } from '../src/lib/style-text.js';
 import { getDefault, getEnumValues, isBoolean, isRequired } from '../src/lib/zod-utils.js';
 import { Logger } from '../src/logger.js';
@@ -35,16 +36,22 @@ import { Logger } from '../src/logger.js';
 // Exit cleanly if the program we pipe to exits abruptly
 process.stdout.on('error', (error) => {
   if (error.code === 'EPIPE') {
-    process.exit(0);
+    exit(0);
   }
 });
+
+process.on('SIGINT', () => exit(130));
+process.on('SIGTERM', () => exit(143));
 
 // Right now, this is the only way to do this properly
 // cliparse doesn't allow unknown options/arguments
 if (process.argv[2] === 'curl') {
-  curl().catch(() => process.exit(1));
+  curl().then(
+    () => exit(0),
+    () => exit(1),
+  );
 } else {
-  run().catch(() => process.exit(1));
+  run().catch(() => exit(1));
 }
 
 async function run() {
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -18,6 +18,7 @@ import { EXPERIMENTAL_FEATURES, getFeatures } from '../src/config/features.js';
 import { cliparse } from '../src/lib/cliparse-patched.js';
 import { styleText } from '../src/lib/style-text.js';
 import { getDefault, getEnumValues, isBoolean, isRequired } from '../src/lib/zod-utils.js';
+import { Logger } from '../src/logger.js';
 
 /**
  * @typedef {import('../src/lib/define-command.types.js').CommandDefinition} CommandDefinition
@@ -211,7 +212,7 @@ function convertOption(option) {
     // Log deprecation warning if option is deprecated
     if (option.deprecated) {
       const message = typeof option.deprecated === 'string' ? `, ${option.deprecated}.` : '';
-      console.error(styleText('yellow', `Warning: --${option.name} is deprecated${message}`));
+      Logger.warn(`--${option.name} is deprecated${message}`);
     }
 
     const result = option.schema.safeParse(value);
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -44,6 +44,13 @@ export default [
     },
   },
   {
+    name: 'no-console',
+    files: ['bin/*.js', 'src/**/*.js'],
+    rules: {
+      'no-console': 'error',
+    },
+  },
+  {
     name: 'no-extraneous-dependencies',
     files: ['scripts/**/*.js'],
     rules: {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -51,6 +51,21 @@ export default [
     },
   },
   {
+    name: 'no-process-exit',
+    files: ['bin/*.js', 'src/**/*.js'],
+    ignores: ['src/lib/exit.js'],
+    rules: {
+      'no-restricted-properties': [
+        'error',
+        {
+          object: 'process',
+          property: 'exit',
+          message: 'Use exit() from src/lib/exit.js — it flushes log buffers before terminating.',
+        },
+      ],
+    },
+  },
+  {
     name: 'no-extraneous-dependencies',
     files: ['scripts/**/*.js'],
     rules: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@babel/traverse": "7.28.5",
         "@clevercloud/client": "12.0.0",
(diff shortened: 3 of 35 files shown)


hsablonniere:
@pdesoyres-cc I rebased to ease the review and tests

davlgd:
Tested `dbb42344` against master (`0ed8f5cf`) on macOS, fresh `npm ci` on both sides. Source and CJS tests ran on Node 24.20.0; the generated macOS ARM64 binary embeds Node 24.18.1. Each point says whether it used the live API or a replaced response. File logging does work: running `--version` with a fresh home directory creates `Library/Logs/clever-cloud/clever-tools.2026-09-10.1.log`, and normal runs write JSON lines into it.

**1. The generated CJS bundle and macOS ARM64 binary fail to start.**
`node scripts/bundle-cjs.js 4.11.0 false` then `node build/4.11.0/clever.cjs profile` → `Error: Cannot find module '.../build/4.11.0/lib/worker.js'`. The same bundle on master prints the profile.
`node scripts/build-binary.js 4.11.0` builds fine (exit 0), but running the binary it produces — `./build/4.11.0/macos/clever-tools-4.11.0_macos/clever --version` — fails before printing anything to stdout: `Error: unable to determine transport target for "pino-roll"` at `fixTarget (/snapshot/branch/build/4.11.0/clever.cjs)` → `pino.transport` → `setup`, exit 1. The same binary built from master prints `4.11.0` and exits 0.

**2. A handled error can exit 0.**
`node bin/clever.js env -a doesnotexist123; echo $?` → master 1, branch 0, both printing `[ERROR] There is no linked or targeted application...`.
In that run `process.exit` was not reached: instrumenting it shows only `beforeExit(0)` / `exit(0)` — node exits before the awaited `shutdown()` completes. The outcome depends on whether another handle keeps the event loop alive: with a preloaded `setInterval`, the same command exits 1.
`CLEVER_QUIET=1 node bin/clever.js env -a doesnotexist` → no stdout, no stderr, exit 0.

**3. Crash when the log directory isn't writable.**
`HOME` pointed at a `chmod 555` directory, `clever --version`: the branch prints `4.11.0`, then `RangeError: The value of "targetStart" is out of range. It must be >= 0. Received -2` at `node_modules/thread-stream/index.js:510:11`. master prints `4.11.0` and exits clean. Same branch with a writable directory is fine.

**4. `npm run validate` fails (exit 2).** `lint` and `format:check` pass, `typecheck` doesn't:
`node_modules/thread-stream/index.d.ts(96,73): error TS2694: Namespace '"worker_threads"' has no exported member 'TransferListItem'`. On master, `npm run typecheck` passes.

**5. `Logger.println()` with no argument prints `undefined`.** Live:

```
$ node bin/clever.js applications list
undefined
• Organization 'David Legrand' (user_…) with 265 applications:
undefined
APPLICATION ID  NAME …
```

master prints blank lines there. `grep -rn "Logger.println()" src bin` returns 24 matches.

**6. The deprecation warning is missing.**
`node bin/clever.js env --add-export -a doesnotexist` → master prints ``Warning: --add-export is deprecated, use `--format shell` instead.``; the branch omits it (the app-not-found error still prints). With `CLEVER_VERBOSE=1` the branch does print `[WARN] --add-export is deprecated…`.
With `open` stubbed to reject and simulated API responses, master prints the browser-opening warning at `login.command.js:60` and the branch omits it; both still print the manual fallback instruction.

**7. Errors that used to be matched on `responseBody` / `response`.**
- `emails add <the account's own primary address>`, live: master `[ERROR] This address already belongs to your account`, branch `[ERROR] This email address already belongs to you [101]`. The specific message for 101 is lost.
- `profile` with `GET /v2/self` and `GET /v2/self/tokens/current` replaced by HTTP 401: master `[ERROR] Your token is invalid or has expired, use clever login command`, branch `[ERROR] You're not logged in, use clever login command…`.
- `domain overview -F json`, live except one `/vhosts` response replaced by HTTP 403 (one app out of 265): master warns on stderr and still prints the full inventory (159 KB of JSON, exit 0); the branch writes `[ERROR] Forbidden domains [4003]` to stderr and produces no JSON on stdout. Without the injected 403 both outputs are identical.

**8. `activity` warns on known states.**

```
$ node bin/clever.js activity --app app_98e2ee70-…
⚠ Unknown deployment state: FAIL
2026-09-10T17:41:47+02:00  deployment_e3b58a1f-…  FAIL  DEPLOY  84d81eb4  Git
⚠ Unknown deployment state: OK
…
```

Every row of the output I got carried one, including `OK` and `FAIL`. Looks like the `Logger.printWarning` at `activity.command.js:31`, first statement of `getColoredState`.

The eight issues were also reproduced from separate source snapshots with fresh `npm ci` installs; the API-error and browser-failure checks used stubs.

PR #1095, opened 2026-05-19
feat(drain): add `clever drain check` command

KannarFr:
## Summary

- Add new `clever drain check <drain-id>` command that probes the drain recipient server-side via `POST /v4/drains/.../check` and prints the resulting `ProbeResult` (`ok`/`code`/`message`).
- Goal: give users a CLI-first way to diagnose drain connectivity issues without having to inspect drain status or recreate the drain.
- Output: human-readable by default (`Probe: OK|FAILED (code)` + message), `--format json` returns the raw `ProbeResult`.

Example:
```
$ clever drain check 01H...
Probe: FAILED (bad-credentials)
POST https://… returned 401 — check API key or credentials
```

## Test plan

- [x] `clever drain check --help` shows the command
- [x] `clever drain check <drain-id>` against a healthy drain returns `Probe: OK`
- [x] `clever drain check <drain-id>` against a misconfigured drain (bad URL / bad creds) returns `Probe: FAILED` with a useful `code`
- [x] `clever drain check <drain-id> --format json` prints the raw ProbeResult
- [x] Works with `--app`, `--alias`, and `--addon` resolvers
- [x] `npm run validate` passes (lint, prettier, tsc, docs:check)

### DIFF
--- a/skills/clever-tools/references/full-documentation.md
+++ b/skills/clever-tools/references/full-documentation.md
@@ -1056,6 +1056,30 @@ clever drain [options]
 -F, --format <format>          Output format (human, json) (default: human)
 ```
 
+### drain check
+
+**Description:** Check that a drain's recipient is reachable and accepts deliveries
+
+**Since:** unreleased
+
+**Usage**
+```
+clever drain check <drain-id> [options]
+```
+
+**Arguments**
+```
+drain-id                       Drain ID
+```
+
+**Options**
+```
+    --addon <addon-id>         Add-on ID or real ID
+-a, --alias <alias>            Short name for the application
+    --app <app-id|app-name>    Application to manage by its ID (or name, if unambiguous)
+-F, --format <format>          Output format (human, json) (default: human)
+```
+
 ### drain create
 
 **Description:** Create a drain
--- a/src/clever-client/drains.js
+++ b/src/clever-client/drains.js
@@ -108,3 +108,20 @@ export function enableDrain(params) {
     // no body
   });
 }
+
+/**
+ * POST /v4/drains/organisations/{ownerId}/resources/{resourceId}/drains/{drainId}/check
+ * @param {Object} params
+ * @param {String} params.ownerId
+ * @param {String} params.resourceId
+ * @param {String} params.drainId
+ */
+export function checkDrain(params) {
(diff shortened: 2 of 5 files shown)


hsablonniere:
Pushed a few adjustments on top of this branch — kept you as `Co-authored-by` on the commit. Here's what changed and the reasoning:

### Lighter, forward-compatible output
- `--format json` now returns a controlled `{ ok, message }` instead of the raw probe object. Dropped `code` from the structured output: it's already summarized in `message`, and it's meant to be superseded by a real HTTP status under `response.statusCode` in a later iteration. The point is to **not freeze a JSON shape we'd have to break next**.
- Human output: `Probe: OK|FAILED` + message (removed the `(code)` suffix).

### `since: 4.11.0` → `unreleased`
- Follows the repo convention — release-please rewrites `unreleased` to the actual version at release time. Hardcoding `4.11.0` is a risky guess (next release could be a patch/major). Docs regenerated accordingly.

### No `throw` on a KO probe
- The command no longer exits non-zero when `probe.ok === false`. Getting a ProbeResult means the check **ran successfully**; whether the drain is healthy is business data carried in the output, orthogonal to the command's success. Real failures (drain not found, API down, auth) still exit non-zero upstream via `sendToApi`.
- Bonus: a CI-friendly exit signal can be added later as an opt-in `--strict` flag **without breaking** the default exit code — whereas removing a throw afterwards *would* have been breaking.

### Description
- `Probe a drain recipient to diagnose connectivity` → `Check that a drain's recipient is reachable and accepts deliveries`. `connectivity` was misleading: this isn't a network/TCP ping but a real HTTP POST whose 2xx status we check — i.e. whether the recipient would actually accept a log delivery.

### Next iteration (API side)
The real value — showing the **request actually sent** and the **response received** — lives server-side, since the API is what performs the probe. We're going to prepare an API update to enrich the ProbeResult with `request { method, url, headers, body }`, `response { statusCode, headers, body }` and `durationMs`. The CLI is already wired to display these as soon as they're returned (purely additive, non-breaking). An internal issue will track the API work.

Net: this keeps the PR a **light, non-blocking first iteration**.

hsablonniere (approved):
LGTM 👍

PR #1097, opened 2026-06-16
feat(global): add --insecure option to skip TLS certificate verification

wowi42:
## What

Add an opt-in `--insecure` global flag and `CLEVER_INSECURE` environment variable to skip TLS certificate verification.

## Why

Some users are behind corporate proxies that intercept TLS with their own certificates. Certificate verification then fails on every request and the CLI is unusable for them. This gives them an explicit, documented escape hatch.

## How

- `src/initial-setup.js`: when `--insecure` is passed or `CLEVER_INSECURE` is set, it runs before any TLS module loads and sets `NODE_TLS_REJECT_UNAUTHORIZED=0`. That single switch covers API calls (`fetch`/undici) and Git operations (the `https.Agent` used by isomorphic-git).
- `src/commands/curl/curl.command.js`: the spawned `curl` process ignores that env var, so `--insecure` is appended to its args when `CLEVER_INSECURE=1` (skipped if `-k`/`--insecure` is already there).
- `src/commands/global.options.js` + `bin/clever.js`: register the new global option so it parses and shows in help.
- Docs regenerated.

## Notes

- Default behavior stays fully secure. Node still prints its own insecure-TLS warning when enabled.
- Due to existing `clever curl` special-casing, with curl the flag must follow the subcommand (`clever curl --insecure <url>`); `CLEVER_INSECURE=1` works in any position.
- `npm run validate` passes (lint, prettier, typecheck, docs:check).

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -10,6 +10,7 @@ import { globalCommands } from '../src/commands/global.commands.js';
 import {
   colorOption,
   helpOption,
+  insecureOption,
   updateNotifierOption,
   verboseOption,
   versionOption,
@@ -68,6 +69,7 @@ async function run() {
       verbose: verboseOption,
       color: colorOption,
       'update-notifier': updateNotifierOption,
+      insecure: insecureOption,
     },
   };
 
@@ -78,7 +80,12 @@ async function run() {
     name: 'clever',
     description: rootCommandDefinition.description,
     version: pkg.version,
-    options: [convertOption(colorOption), convertOption(updateNotifierOption), convertOption(verboseOption)],
+    options: [
+      convertOption(colorOption),
+      convertOption(updateNotifierOption),
+      convertOption(verboseOption),
+      convertOption(insecureOption),
+    ],
     helpCommand: false,
     commands: sortedCommands,
   });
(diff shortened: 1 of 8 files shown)


hsablonniere:
Thanks a lot for this @wowi42 — the corporate-proxy pain point is real and worth solving.

We'd rather fix it without ever turning TLS verification off, so we went with #1099 instead. The motivating case (a proxy intercepting HTTPS with its own certificate) is almost always one where the proxy/corporate root CA is, or can be, installed at the OS level. #1099 makes the binary trust the OS certificate store (`pkg --options use-system-ca`), and documents `NODE_EXTRA_CA_CERTS=/path/ca.pem` for CAs only available as a file — covering both the binary and npm installs, with verification kept on.

The concern with `--insecure` / `CLEVER_INSECURE` (i.e. `NODE_TLS_REJECT_UNAUTHORIZED=0`) is that it disables verification for the **whole** process — API calls, Git operations and `clever curl` alike — including the transmission of the Clever Cloud token/secret, which then becomes interceptable. That's a large, silent blast radius for what is usually a missing-CA problem.

So closing this in favor of #1099. If you hit a case it doesn't cover (a cert that genuinely can't be trusted via the OS store or a file), please let us know on #1099 — we can reconsider a scoped escape hatch then.

PR #1101, opened 2026-06-24
fix(git): hint at the system git backend when run from a git worktree

### LINKED ISSUE #1096: clever deploy: "Could not find HEAD" when run from a git worktree
## Summary

Running `clever deploy` (or any command that resolves the current commit) from inside a **linked git worktree** fails with:

```
[ERROR] Could not find HEAD.
```

The same repository deploys fine from its main checkout. It looks like clever-tools reads `.git/HEAD` (or otherwise assumes `.git` is a directory) instead of resolving the worktree's `gitdir`.

## Why it happens

In a linked worktree created with `git worktree add`, the top-level `.git` is **a file**, not a directory:

```
$ file .git
.git: ASCII text
$ cat .git
gitdir: /path/to/repo/.git/worktrees/<name>
```

`HEAD` for that worktree lives at `<gitdir>/HEAD` (e.g. `/path/to/repo/.git/worktrees/<name>/HEAD`), not at `<repo>/.git/HEAD`. Resolving the current commit should follow the `gitdir:` pointer (or just shell out to `git rev-parse HEAD` / `git symbolic-ref HEAD`), which works correctly inside a worktree.

## Steps to reproduce

```bash
# from a normal clone that is linked to a Clever Cloud app (.clever.json present)
git worktree add ../myrepo-wt -b some-branch
cd ../myrepo-wt
# commit something here
clever deploy -a <app-alias>
# => [ERROR] Could not find HEAD.
```

## Expected

`clever deploy` resolves the worktree's `HEAD` and force-pushes it to the app's deployment remote, exactly as it does from the main working tree.

## Workaround

Deploy by pushing directly to the app's git remote instead of `clever deploy`:

```bash
git push "ssh://git@push-<zone>.services.clever-cloud.com/<app_id>.git" HEAD:master --force
```

## Environment

- clever-tools: **4.7.1**
- git: 2.54.0
- Node.js: v26.2.0
- OS: Linux

This is increasingly common as git worktrees become a standard workflow (e.g. AI coding agents that isolate work in `.git/worktrees/...`). Thanks!

hsablonniere:
Closes #1096

## Context

`clever deploy` from a linked git worktree (`git worktree add`) fails with a cryptic `[ERROR] Could not find HEAD.`

The default JS git backend (isomorphic-git) resolves `.git` as a directory. In a linked worktree `.git` is a *file* holding a `gitdir:` pointer, and refs live in the shared `commondir` — isomorphic-git supports neither. A real fix would mean routing every operation between the worktree gitdir and the commondir, on a backend we plan to phase out. The system git backend already handles worktrees natively, since it shells out to the installed `git`.

## Proposal

Two commits:

1. **`fix(git)`** — detect linked worktrees in the shared repo accessor and raise an actionable error pointing to the system git backend. The guard covers every JS-backend git operation, so `deploy`, `restart` and `create` all surface the same hint instead of crashing with `Could not find HEAD` / `ENOTDIR`:

   ```
   [ERROR] Linked git worktrees aren't supported by the default JS git backend.
           Enable the system git backend (it uses your installed git):
               clever features enable system-git
   ```

2. **`docs(deploy)`** — list worktrees among the JS backend limitations in the `🧪 Experimental: System git backend` section.

### Design decisions

- `isInsideGitRepo()` keeps returning `true` in a worktree (it bypasses the guard), so `clever create` doesn't wrongly suggest `git init`.
- `clever create`'s working-tree-clean hint is best-effort: it now swallows the worktree error instead of aborting the command.

## How to test

```bash
git worktree add ../wt -b some-branch && cd ../wt
clever deploy                       # actionable hint instead of "Could not find HEAD"
clever features enable system-git
clever deploy                       # deploys from the worktree
```

### DIFF
--- a/src/commands/create/create.command.js
+++ b/src/commands/create/create.command.js
@@ -9,6 +9,7 @@ import { Logger } from '../../logger.js';
 import * as AppConfig from '../../models/app_configuration.js';
 import * as Application from '../../models/application.js';
 import { AVAILABLE_ZONES, listAvailableTypes, listAvailableZones } from '../../models/application.js';
+import { LinkedWorktreeNotSupportedError } from '../../models/git-isomorphic.js';
 import { Git } from '../../models/git.js';
 import { aliasCreationOption, humanJsonOutputFormatOption, orgaIdOrNameOption } from '../global.options.js';
 
@@ -54,7 +55,13 @@ async function displayAppCreation(app, alias, github, taskCommand) {
       Logger.println(`    ${shellCommand('git commit -m "Initial commit"')}`);
       Logger.println();
     } else {
-      const isClean = await git.isGitWorkingDirectoryClean();
+      // Best-effort hint: if we can't tell because of a linked worktree on the JS backend, don't nag
+      const isClean = await git.isGitWorkingDirectoryClean().catch((error) => {
+        if (error instanceof LinkedWorktreeNotSupportedError) {
+          return true;
+        }
+        throw error;
+      });
       if (!isClean) {
         Logger.println(`  ${styleText('yellow', '!')} Commit your changes first:`);
         Logger.println(`    ${shellCommand('git add .')}`);
--- a/src/commands/deploy/deploy.docs.md
+++ b/src/commands/deploy/deploy.docs.md
@@ -28,6 +28,7 @@ Clever Tools uses a current JS implementation for git operations. This works wit
 * **HTTP-only**: cannot use SSH-based git protocols
 * **Slow performance** on repositories with rewritten history (rebases, squashes)
 * **Connection timeouts** on large repositories or when pushing big files, due to HTTP-based transfers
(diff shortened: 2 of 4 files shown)


src/commands/create/create.command.js
```diff
@@ -54,7 +54,8 @@ async function displayAppCreation(app, alias, github, taskCommand) {
       Logger.println(`    ${shellCommand('git commit -m "Initial commit"')}`);
       Logger.println();
     } else {
-      const isClean = await git.isGitWorkingDirectoryClean();
+      // Best-effort hint: if we can't tell (e.g. linked worktree on the JS backend), don't nag
+      const isClean = await git.isGitWorkingDirectoryClean().catch(() => true);
```
pdesoyres-cc: fix: we should narrow the error to  `LinkedWorktreeNotSupportedError` otherwise we may hide other errors like corrupted repo, permission error, etc...

pdesoyres-cc (changes requested):

pdesoyres-cc (approved):

PR #1102, opened 2026-06-24
fix(prompts): surface a clear error when prompts run without a TTY

### LINKED ISSUE #1092: Raw-mode prompt commands fail silently with exit 1 in non-TTY contexts
## Summary

Commands that prompt the user via `selectAnswer` or `promptCheckbox` (raw-mode `@inquirer/prompts`) exit with code 1 and **no error message** when run without a TTY (CI pipelines, scripts, subshells without a controlling terminal).

## Reproduction

With three or more profiles configured:

```bash
clever profile switch < /dev/null
```

Observed:

- exit code: `1`
- stdout: a half-rendered prompt (`? Select a profile: (Use arrow keys)` plus the choice list)
- stderr: *empty*

The user gets no diagnostic, no remediation hint, and no indication that the command requires an interactive terminal.

## Root cause

`@inquirer/prompts.select()` needs raw-mode stdin (`process.stdin.setRawMode(true)`), which only exists on TTYs. When stdin is a pipe, Inquirer reads EOF immediately and throws an `ExitPromptError` (its "user cancelled" path).

`exitOnPromptError` in `src/lib/prompts.js:53–58` then calls `process.exit(1)` silently:

```js
function exitOnPromptError(error) {
  if (error instanceof Error && error.name === 'ExitPromptError') {
    process.exit(1);
  }
  throw error;
}
```

## Affected commands

Every command using `selectAnswer` or `promptCheckbox` from `src/lib/prompts.js`:

- `profile switch` (when 3+ profiles, no `--alias`)
- `ssh` (when multiple instances; partially guarded today by an isTTY check at `src/commands/ssh/ssh.command.js:48` — that path *does* throw, but only when stdin closes before Inquirer renders)
- `oauth-consumers create` / `oauth-consumers update` (rights checkbox)
- `k8s cluster-version-update` / `operator version-update` (version select)

`confirm` / `ask` / `password` / `input` are unaffected — they fall back to readline and work over pipes.

## Existing precedent

`src/commands/ssh/ssh.command.js:48` already gates one specific case explicitly:

\`\`\`js
if (process.stdout.isTTY) {
  // prompt
} else {
  throw new Error(/* explicit message */);
}
\`\`\`

The same idea should apply uniformly to \`selectAnswer\` and \`promptCheckbox\`.

## Proposed fix

Two options, ideally combined:

1. **Central guard in \`src/lib/prompts.js\`** — \`selectAnswer\` and \`promptCheckbox\` check \`process.stdout.isTTY\` (or \`process.stdin.isTTY\`) up front and throw a clear error: *"This command requires an interactive terminal."*
2. **Per-command override** where a remediation hint exists. For \`profile switch\`: *"Cannot prompt for profile selection in non-interactive shells. Use \`--alias <name>\` to specify a profile directly."*

hsablonniere:
Closes #1092

## Context

Commands that prompt via `selectAnswer` or `promptCheckbox` (raw-mode `@inquirer/prompts`) exited with code `1` and **no error message** when run without a controlling terminal (CI, scripts, `< /dev/null`).

- Inquirer reads EOF on a non-TTY stdin and throws `ExitPromptError`
- `exitOnPromptError` turns that into a silent `process.exit(1)` — bypassing the normal `Logger.error` path
- result: exit 1, empty stderr, a half-rendered prompt left on screen

## Proposal

Guard `selectAnswer` / `promptCheckbox` up front on `process.stdin.isTTY` and throw an explicit error instead of letting Inquirer fail silently.

- `This command requires an interactive terminal.`
- with an optional per-command remediation hint, e.g. `Use --alias <name> to select a profile directly.`

Hints are added only where a non-interactive option exists:

| command | hint |
|---|---|
| `profile switch` | `--alias <name>` |
| `oauth-consumers create/update` | `--rights <list>` |
| `k8s version update` | `--target <version>` |
| `operator version update` (keycloak/metabase/otoroshi) | `--target <version>` |

`ssh` keeps its specific message (multiple instances) and has no hint (no selection option).

### Design decisions

- **Ctrl+C stays silent.** Ctrl+C and non-TTY both produce the same `ExitPromptError`, so they can't be told apart *after* the fact. The guard runs *before* the prompt and only triggers when `stdin` is not a TTY — so an interactive Ctrl+C still flows through the unchanged `exitOnPromptError` (silent exit), no regression.
- **`ssh` guard fix.** It tested `process.stdout.isTTY`; the failing case (`clever ssh < /dev/null` in a terminal) has a TTY stdout but a non-TTY stdin, so it slipped through. Now tests `stdin`.

## Commits

Split for a clean changelog (1 fix + 4 features; the refactor is internal):

- `fix(prompts)` — the guard + ssh fix
- `refactor(prompts)` — optional `nonInteractiveHint` argument
- `feat(profile)` / `feat(oauth-consumers)` / `feat(k8s)` / `feat(operators)` — per-command hints

## How to test

```bash
# non-TTY: clear message + exit 1 (was: silent)
node bin/clever.js profile switch < /dev/null

# Ctrl+C in an interactive terminal: still exits silently (no stacktrace)
node bin/clever.js profile switch   # then press Ctrl+C

# nominal: arrows + Enter still switches profile
node bin/clever.js profile switch
```

### DIFF
--- a/src/commands/profile/profile.switch.command.js
+++ b/src/commands/profile/profile.switch.command.js
@@ -75,6 +75,10 @@ async function resolveTargetProfile(profiles, requestedAlias) {
     name: formatProfile(profile),
     value: profile.alias,
   }));
-  const selectedAlias = await selectAnswer('Select a profile:', choices);
+  const selectedAlias = await selectAnswer(
+    'Select a profile:',
+    choices,
+    'Use --alias <name> to select a profile directly.',
+  );
   return profiles.find((profile) => profile.alias === selectedAlias);
 }
--- a/src/commands/ssh/ssh.command.js
+++ b/src/commands/ssh/ssh.command.js
@@ -45,7 +45,7 @@ export const sshCommand = defineCommand({
     let sshTarget;
     if (instances.length === 1) {
       sshTarget = instances[0].id;
-    } else if (process.stdout.isTTY) {
+    } else if (process.stdin.isTTY) {
       const choices = instances
         .sort((a, b) => a.instanceNumber - b.instanceNumber)
         .map((inst) => ({
--- a/src/lib/k8s.js
+++ b/src/lib/k8s.js
@@ -503,6 +503,7 @@ export async function k8sUpdateVersion(orgIdOrName, clusterIdOrName, askedVersio
     (await selectAnswer(
       `Which version do you want to update ${styleText('blue', name)} to, current is ${styleText('blue', versions.installed)}?`,
       [...versions.available].reverse(),
+      'Use --target <version> to update directly.',
     ));
 
   if (!versions.available.includes(targetVersion)) {
--- a/src/lib/operator-commands.js
+++ b/src/lib/operator-commands.js
@@ -77,6 +77,7 @@ export async function operatorUpdateVersion(provider, askedVersion, addonIdOrNam
     (await selectAnswer(
       `Which version do you want to update ${styleText('blue', name)} to, current is ${styleText('blue', versions.installed)}?`,
       versions.available.reverse(),
+      'Use --target <version> to update directly.',
     ));
 
   if (!versions.available.includes(targetVersion)) {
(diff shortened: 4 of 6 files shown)


pdesoyres-cc (approved):
I like this

PR #1103, opened 2026-06-24
fix(cli): stop printing valid positional args as undefined errors

### LINKED ISSUE #1087: Confusing error output: "<arg-name>: undefined" for valid positional args when a sibling fails
## Context

When a command takes multiple positional arguments and one of them fails Zod validation, the error output also lists the **valid** siblings as `undefined`, which is confusing.

## Reproduction

```
$ clever k8s nodegroups create kubernetes_01XXXXXXXXXXXXXXXXXXXXXXXX nope BadFormat
cluster-id|cluster-name: undefined
nodegroup-name: undefined
flavor:count: Expected format: <flavor>:<count>
```

Here `kubernetes_01XXXXXXXXXXXXXXXXXXXXXXXX` and `nope` are perfectly valid as `cluster-id|cluster-name` and `nodegroup-name`. Only `flavor:count` is invalid, but the user reads three errors.

Same pattern reproduces on any `defineCommand` with `args: [arg1, arg2, arg3]` when only one fails.

## Root cause

`cliparse@0.5.0` (`node_modules/cliparse/src/argument.js:34-55`) collects parse results for all arguments, and as soon as one is in error, returns the **whole list** wrapped as `parsers.error(results)`. Then `node_modules/cliparse/src/cli.js:54-65` (`cli.displayErrors`) iterates over every entry and prints `arg.argument.name + ': ' + arg.error`. Since the success entries have no `.error` field, they print `: undefined`.

## Suggested fix

Patch `cliparseCli.displayErrors` in `src/lib/cliparse-patched.js` to filter out entries where `result.error == null` before delegating to the original implementation:

```js
import cliparseCli from 'cliparse/src/cli.js';
const original = cliparseCli.displayErrors;
cliparseCli.displayErrors = (errors, cliApp, givenArgs) => {
  const filtered = {
    args: (errors.args ?? []).filter((r) => r.error != null),
    options: errors.options ?? {},
  };
  return original(filtered, cliApp, givenArgs);
};
```

Verified locally: real errors still display, valid args no longer leak as `undefined`, and the case where every positional is missing keeps printing `<arg-name>: missing value` for each.

## Scope

Out of scope of the current 4.9 release PR (Kubernetes commands), so opening this as a follow-up.

hsablonniere:
Closes #1087

## Context
When a command takes several positional args and only one fails its parser, `cliparse@0.5.0` keeps the **whole** result list as the error payload (`argument.parseList`, `src/argument.js:53`). `cli.displayErrors` then prints the valid siblings as `<arg-name>: undefined`, so the user reads several errors when only one is real.

```
$ clever k8s nodegroups create kubernetes_01XXX nope BadFormat
cluster-id|cluster-name: undefined
nodegroup-name: undefined
flavor:count: Expected format: <flavor>:<count>
```

## Proposal
Patch `argument.parseList` in `src/lib/cliparse-patched.js` (the existing dedicated patch module) to keep only the entries that actually failed in its error payload.

### Design decisions
- Fixed at the **root** (`parseList`) rather than at the display layer: the error payload no longer carries successful results, instead of cleaning them up just before printing. The error array has a single consumer (`displayErrors`), so the effect is identical, but the fix sits where the wrong data is produced.
- Wrapper only rewrites the **error** branch — the success path (args passed to the handler) is untouched.
- Missing-value errors carry an `.error` (`"missing value"`), so a fully empty arg list still reports each arg.
- Kept local rather than patching `cliparse` upstream, consistent with the other patches in this module.

## How to test
`k8s` commands are behind a feature flag. With it enabled:

```
clever k8s nodegroups create kubernetes_01XXX nope BadFormat
# => flavor:count: Expected format: <flavor>:<count>   (single error)
```

Cases verified:
- 1 invalid arg out of 3 → only the invalid one is shown
- middle arg invalid → only that one
- all args missing → each reported as `missing value`
- option error (e.g. `--min notanumber`) → preserved
- all valid → handler runs normally with the parsed args

### DIFF
--- a/src/lib/cliparse-patched.js
+++ b/src/lib/cliparse-patched.js
@@ -1,4 +1,5 @@
 import cliparseOriginal from 'cliparse';
+import cliparseArgumentModule from 'cliparse/src/argument.js';
 import cliparseCommandModule from 'cliparse/src/command.js';
 import semver from 'semver';
 import pkg from '../../package.json' with { type: 'json' };
@@ -31,6 +32,23 @@ cliparseOriginal.command = function (name, options, commandFunction) {
     });
   });
   return command;
+};
+
+// Patch cliparse.argument.parseList to drop the parse results that succeeded
+// from its error payload. When several positional args are given and only one
+// fails its parser, parseList returns the *whole* result list (successes
+// included) as the error. Downstream, displayErrors prints the valid siblings
+// as `<arg-name>: undefined` (they carry an `.argument` but no `.error`).
+// We keep only the entries that actually failed, fixing the issue at its root.
+// The success path is untouched (we only rewrite the error array), and
+// missing-value errors carry an `.error` ("missing value") so they are kept.
+const originalParseList = cliparseArgumentModule.parseList;
+cliparseArgumentModule.parseList = function (args, providedArguments) {
+  const result = originalParseList(args, providedArguments);
+  if (Array.isArray(result.error)) {
+    return cliparseOriginal.parsers.error(result.error.filter((entry) => entry.error != null));
+  }
+  return result;
 };
 
 /**


pdesoyres-cc (approved):
LGTM

PR #1104, opened 2026-06-24
feat(accesslogs): restore CLF output format

### LINKED ISSUE #909: Support CLF format on accesslogs
was supported on previous version

https://www.clever-cloud.com/blog/features/2020/04/09/access-logs-clever-tools/
https://en.wikipedia.org/wiki/Common_Log_Format
https://github.com/npm/common-log-object/blob/master/index.js

hsablonniere:
Closes #909

## Context

The Common Log Format was dropped during the Warp10 → v4 API migration (commit `619095d9`, early 2024), which reduced the available formats to `human`, `json` and `json-stream`. CLF was supported by the previous CLI and is still advertised in the public docs, so this is a regression (reopened via #909).

## Proposal

Re-add `--format clf` on `clever accesslogs`, backed by a dedicated formatter (`src/lib/access-logs-clf.js`) mapping the v4 access log shape to a CLF line:

```
203.0.113.7 - - [24/Jun/2026:08:05:43 +0000] "GET /apache_pb.gif" 200 2326
```

### Design decisions

- **Dedicated `accessLogsFormatOption`** instead of extending the shared `logsFormatOption` — that option is also used by `clever logs` (runtime logs), where an HTTP-only format is meaningless.
- **Request line is `method path` (no `HTTP/x.y`)** — the protocol version is absent from the v4 payload. The protocol token is optional in CLF parsers, and a `-` placeholder would actually break structured parsing (verified against the grok `COMMONAPACHELOG` pattern and GoAccess, both fall back to an unparsed request with `-`).
- **UTC `+0000`** for the timestamp — deterministic, no `clf-date` dependency re-added (custom helper instead, in line with the project trend of dropping small deps).
- **`clf` streams line by line** like `human`/`json-stream` → not subject to the `--before` requirement that `json` has (the latter buffers a JSON array that must be closed).
- TCP redirections / early-cut connections have no `http` section → skipped (same guard as `human`).

## How to test

```bash
clever accesslogs --format clf            # live tail, CLF lines
clever accesslogs --format clf --before 1h  # bounded fetch
```

Out of scope: the `extended` format (also removed in 2024) is not reintroduced — #909 only asks for CLF.

### DIFF
--- a/src/lib/access-logs-clf.js
+++ b/src/lib/access-logs-clf.js
@@ -1,0 +1,52 @@
+const CLF_MONTHS = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+
+/**
+ * Format a `Date` as a Common Log Format timestamp in UTC, e.g. `10/Oct/2000:13:55:36 +0000`.
+ * @param {Date} date
+ * @returns {string}
+ */
+function formatClfDate(date) {
+  const pad = (n) => String(n).padStart(2, '0');
+  const day = pad(date.getUTCDate());
+  const month = CLF_MONTHS[date.getUTCMonth()];
+  const year = date.getUTCFullYear();
+  const hours = pad(date.getUTCHours());
+  const minutes = pad(date.getUTCMinutes());
+  const seconds = pad(date.getUTCSeconds());
+  return `${day}/${month}/${year}:${hours}:${minutes}:${seconds} +0000`;
+}
+
+/**
+ * Escape a value so it can be safely embedded inside a double-quoted CLF field.
+ * Backslashes are escaped first, then double-quotes, otherwise a `"` would be escaped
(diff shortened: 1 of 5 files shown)


src/lib/access-logs-clf.js
```diff
@@ -0,0 +1,40 @@
...
+ * @see https://en.wikipedia.org/wiki/Common_Log_Format
+ * @param {object} log an HTTP access log (the `http` section must be present)
+ * @returns {string}
+ */
+export function formatClf(log) {
+  const host = log.source.ip;
+  const ident = '-';
+  const authuser = '-';
+  const date = formatClfDate(log.date);
+  const request = `${log.http.request.method} ${log.http.request.path}`;
```
pdesoyres-cc: fix: we should escape the double-quote character from path... just to be sure. Otherwise, it may break the quoted clf field and shift every subsequent column.

pdesoyres-cc (dismissed):
Well done Hubert. Got one feedback

florian-sanders-cc (approved):

PR #1105, opened 2026-06-24
feat(commands): automate since version field for new commands

### LINKED ISSUE #1039: Automate `since` version field for new commands at release time
## Context

When adding a new command, contributors must specify a `since` field indicating which version introduced the command. This field is used in documentation generation (see `scripts/lib/docs-reference.js`).

```javascript
export const myCommand = defineCommand({
  description: 'What this command does',
  since: '3.14.0',  // ← Problem: we don't know this yet
  // ...
});
```

**The problem**: At commit time, we don't know which version the command will ship in. The version is determined later by release-please based on accumulated commits (feat → minor, fix → patch, etc.).

This leads to:
- Guessing the version number
- Potential inaccuracies in documentation
- Friction for contributors

## Proposal

Make the `since` field optional during development and automatically populate it at release time.

### Option A: Inject in release-please PR (Recommended)

1. Make `since` optional in `defineCommand`
2. Add a script that runs in the `release-please.yml` workflow
3. The script detects commands without `since` and adds them to the release PR with the upcoming version
4. After merge, `master` always has accurate version info

**Pros**: Source code reflects reality after release, integrates well with existing workflow
**Cons**: Slightly more complex CI setup

### Option B: Inject at build/publish time

1. Make `since` optional
2. During doc generation, use the current version from `package.json` for commands without `since`

**Pros**: Simpler implementation
**Cons**: Source code never contains the actual version

### Option C: Use a placeholder value

1. Use `since: null` or `since: 'unreleased'` for new commands
2. Release script replaces placeholders with the actual version

**Pros**: Explicit marker for unreleased commands
**Cons**: Similar to Option A but with extra placeholder handling

## Implementation considerations

- Update `defineCommand` type to make `since` optional
- Handle missing `since` gracefully in doc generation (show "Unreleased" or omit the badge)
- Add validation in CI to ensure all commands have `since` after release

hsablonniere:
Closes #1039

## Context

The `since` field records the version a command shipped in, but that version is unknown at commit time — release-please computes it later from accumulated commits (feat → minor, fix → patch). Contributors had to guess the number or leave an informal `since: 'unreleased'` string (which already crept into `drain check`).

## Proposal

Make `since` an explicit, type-checked contract and fill the real version automatically at release time.

- **Required, `version | null`** — `since` is no longer optional. `null` means "not yet released"; a missing value is now a TypeScript error rather than a silent omission. A real semver value is still type-checked via the template literal.
- **Doc generation** — `since === null` renders as `Unreleased` (clean badge) instead of the previous `Since undefined`.
- **Release automation** — when release-please opens its release PR, a workflow step reads the bumped version from `package.json`, rewrites every `since: null` to that version (`scripts/resolve-since.js`), regenerates the docs, and pushes it back onto the PR. After merge, `master` always has accurate version info.

### Design decisions

- **`null` over a `'unreleased'` placeholder** — the original idea (issue option C) was a magic string. `null` is unambiguous to detect (`=== null`), can't be confused with a real version, and being *required* prevents silent omission. This was the conclusion of the discussion on the issue.
- **Resolve in the release PR (option A)** over resolving at doc-gen time (option B) — keeps the source as the source of truth and avoids mis-dating a command if a `null` survives several releases.

## How to review

1. `src/lib/define-command.types.d.ts` — the contract change (`since: version | null`, required).
2. `scripts/lib/docs-reference.js` + `scripts/lib/docs-llm.js` — the `Unreleased` guard.
3. `scripts/resolve-since.js` — the rewrite script (validates semver, replaces `since: null`).
4. `.github/workflows/release-please.yml` — the release-time step.
5. `drain check` + regenerated docs — migration of the existing `'unreleased'` placeholder.

## Notes

- `npm run docs` hits the public CC API (zones/products) — already the case for `docs:check`, so the release job has the same dependency.
- `code-quality.yml` ignores `release-please--**`, so docs are regenerated in the release step to keep `master` consistent after merge.
- No infinite loop: pushing to the release branch doesn't trigger the workflow (trigger = push to `master`).

### DIFF
--- a/scripts/resolve-since.js
+++ b/scripts/resolve-since.js
@@ -1,0 +1,238 @@
+#!/usr/bin/env node
+//
+// Resolve the `since` marker of every command definition to a concrete version.
+//
+// New commands are committed with `since: null` because the version they ship in
+// is only known at release time (release-please computes it from accumulated
+// commits). This script, run from the release workflow against the release PR
+// branch, rewrites those markers to the upcoming version.
+//
+// It parses each `*.command.js` file with the TypeScript compiler API (AST) rather
+// than regexes, locates the object literal passed to `defineCommand({ ... })`, and
+// resolves its `since` property in two cases:
+//   - `since: null`  -> rewrite the value to `since: '<version>'`.
+//   - `since` absent from the literal:
+//       * if the literal spreads another definition (`...x`), `since` is inherited:
+//         leave the file untouched and do not report it.
+//       * otherwise the property was simply forgotten on a new command: insert
+//         `since: '<version>'` into the literal.
+// An already-set `since: '<string>'` is left strictly intact. Any other value
+// (neither `null` nor a string literal) is a hard error citing the file.
+//
+// Rewrites are surgical: only the touched property changes (via AST node offsets),
+// the rest of the file is preserved byte for byte (no reformatting).
+//
+// USAGE: resolve-since.js <version>
+//
+// ARGUMENTS:
+//   version         The upcoming release version (semver, e.g. 3.14.0)
+//
+// EXAMPLES:
+//   resolve-since.js 3.14.0
+//
+import fs from 'node:fs/promises';
+import { globSync } from 'tinyglobby';
+import ts from 'typescript';
+import { runCommand } from './lib/command.js';
+
+const SEMVER_REGEX = /^\d+\.\d+\.\d+$/;
+
+/**
+ * @typedef {Object} SinceResult
+ * @property {'resolved'|'inserted'|'unchanged'|'skip'} action - what was done to the file.
+ * @property {string} content - the resulting file content (equal to the input when nothing changed).
+ * @property {string} [reason] - why the file was skipped (only set when action is 'skip').
+ */
+
+runCommand(async () => {
+  const version = process.argv[2];
+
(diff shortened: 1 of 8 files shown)


scripts/resolve-since.js
```diff
@@ -0,0 +1,60 @@
...
+const SINCE_NULL_REGEX = /\bsince:\s*null\b/g;
+
+runCommand(async () => {
+  const version = process.argv[2];
+
+  if (version == null || version === '') {
+    throw new ArgumentError('version');
+  }
+  if (!SEMVER_REGEX.test(version)) {
+    throw new ArgumentError(`Invalid version "${version}", expected semver (e.g. 3.14.0)`);
```
pdesoyres-cc: fix: do not use `ArgumentError` here. Prefer using `Error` directly

pdesoyres-cc (changes requested):
NIce addition. My feedbacks:

- the `defineCommand` is typed in `define-command.types.d.ts` with the non-undefined `since` guard. But `tsconfig.json` doesn't include the commands. Today, the guard is not applied.
- `resolve-since.js` could use an AST approach which mayu permit  to also handle the absence of `since` property.

pdesoyres-cc (approved):

PR #1106, opened 2026-06-24
chore(node): upgrade to Node.js 24

### LINKED ISSUE #1013: Upgrade to Node.js 24
## Context

Node.js 24 is now supported by `@yao-pkg/pkg` (since version 3.5.25 of `@yao-pkg/pkg-fetch`).

## Goal

Upgrade the project to use Node.js 24 instead of Node.js 22.

## Tasks

- [ ] Update `engines.node` in `package.json`
- [ ] Update `pkg.node` in `package.json`
- [ ] Update `.mise.toml`
- [ ] Update CI if necessary
- [ ] Test pkg builds on all platforms

hsablonniere:
Closes #1013

## Context

Node.js 24 is now supported by `@yao-pkg/pkg`. The project currently runs on Node.js 22 and pins `@yao-pkg/pkg@6.10.1`, which only resolves `@yao-pkg/pkg-fetch@3.5.30` — and that build ships a single v24 base binary (`24.11.0`).

## Proposal

Bump the runtime to Node.js **24.18.1** (Active LTS), and bump **`@yao-pkg/pkg` 6.10.1 → 6.22.0** so the toolchain can fetch it. `pkg@6.22.0` pulls `pkg-fetch@3.6.5`, whose `patches.json` ships `v24.18.1` as its v24 entry — i.e. what `pkg` fetches for the `node24-*` targets. Staying on `pkg@6.10.1` would lock us to `24.11.0`.

- `@yao-pkg/pkg`: `6.10.1` → `6.22.0` (pulls `pkg-fetch@3.6.5`)
- `engines.node`: `>=22` → `>=24`
- `volta.node` / `mise.toml`: `22.17.0` → `24.18.1`
- `docs/README.md`: system-CA note now mentions `Node.js >= 24`

### ⚠ Breaking change → next release is a major (5.0.0)

Same as the Node 18 → 22 bump, which shipped as **4.0.0**. Bumping the minimum Node version is breaking for users who install via npm.

- **`npm install` users**: now require Node.js 24+.
- **Binary users**: unaffected — Node is embedded in the executable.

The commit is `feat(node)!:` with a `BREAKING CHANGE:` footer so release-please produces the major bump (mirrors `chore!: prepare 4.0.0`).

### Why not 25 or 26

- **25**: odd → never LTS, already EOL, never built by `pkg-fetch`.
- **26**: released May 2026 but still *Current* — only becomes LTS in October 2026. Too early for a distributed binary.
- **24**: Active LTS (Krypton) → the safe target, matching the issue.

### Notes

- CI needs no change: `.github/actions/setup-node` reads `node-version-file: package.json` (i.e. `volta.node`).
- The build target is derived from `volta.node` in `scripts/lib/build-binary.js` (`node${major}-${platform}-${arch}`), so the major bump flows through automatically.
- A newer `pkg-fetch@3.6.6` already ships a `v24.20.0` base binary, but no `@yao-pkg/pkg` release depends on it yet — `24.18.1` is the highest node24 build actually reachable through `pkg` today.

## How to test

1. `mise install` (or `volta install`) → picks up Node 24.18.1
2. `npm ci && npm test`
3. `npm run build:binary` on each platform → confirms `pkg` fetches the Node 24.18.1 base binary and produces working executables (Linux / macOS / Windows)

> Validated locally on Linux: `bundle-cjs` + `build-binary` produce a working executable embedding **Node 24.18.1** (`pkg@6.22.0` / `pkg-fetch@3.6.5`).

## Out of scope — deferred until full e2e coverage

This PR is a pure runtime bump. The "now that we're on Node 24, let's use it" refactors are **deliberately deferred until a complete e2e test suite is in place**, so we can refactor safely (they touch runtime behavior across many commands). To be tracked as follow-up issues/PRs:

- Drop `lodash` in favor of native equivalents (`structuredClone`, `Object.groupBy`, `Array` methods, optional chaining).
- `using` / `await using` for resources currently closed via `try/finally`.
- `RegExp.escape` where regex patterns are built from dynamic input.
- `URLPattern`, native type stripping — only if a concrete need shows up.


pdesoyres-cc (dismissed):
Thanks Hubert.

My feedback:

- branch is not rebased on master
- there are two commits but both update the versions of node AND yao pkg.
- the CI doesn't pass

pdesoyres-cc (approved):
LGTM

PR #1107, opened 2026-06-24
feat(proxy): honor http_proxy/https_proxy environment variables

### LINKED ISSUE #1050: Proxy Support?
Is running the clever cli behind a corporate proxy (with auth) supported?

It doesn't seem to take the http(s)_proxy environment variables into consideration.

hsablonniere:
Closes #1050

## Context

On corporate networks where all outgoing traffic must go through an authenticating HTTP proxy, `clever` failed with `ECONNREFUSED`. Unlike `curl`, `kubectl` or `s3cmd`, Node's global `fetch()` **ignores** the `http_proxy`/`https_proxy` environment variables by default — so the CLI never used the proxy.

The reporter found a workaround (`NODE_USE_ENV_PROXY=1`), but that requires users to know a magic variable, only works on recent Node, and emits an experimental warning on some of the versions where it does. We can do better: behave like the other CLI tools and pick up the proxy automatically.

## Proposal

Install an `EnvHttpProxyAgent` (from `undici`) as the global dispatcher at startup, so **every** request the CLI makes (API calls, update checks…) routes through the proxy:

- triggers on `http_proxy` / `https_proxy` (and the uppercase variants)
- honors `no_proxy` / `NO_PROXY` for exclusions
- supports authenticated proxies via `http://user:password@host:port`
- the dispatcher is **only** swapped when a proxy is actually configured → zero behavior change for users without a proxy

Done in `src/initial-setup.js` because it must run before the first `fetch()`.

### Design decisions

- **`undici` dependency + `setGlobalDispatcher`** rather than the native `NODE_USE_ENV_PROXY=1`: the native variable is read at boot only (can't be enabled from JS at runtime), is gated behind Node ≥ ~22.18, and prints an experimental warning. The `undici` route works at runtime on any supported Node, with no warning, and requires no action from the user.
- `undici` pinned to **7.24.4**, the same major bundled inside Node 24 → the global-dispatcher symbol is shared with the built-in `fetch`.

`undici@7.24.4` requires Node ≥ 20.18.1 and is the same major as the `undici` bundled in Node 24, so the global-dispatcher symbol is shared with the built-in `fetch`.

## Out of scope

- **`system-git` deploys**: with that feature enabled, `clever deploy` delegates to the system `git` binary, which follows its own proxy configuration (`git config http.proxy` / its own reading of `http_proxy`). Not covered here, documented in the README.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -108,6 +108,33 @@ If you switch to the system git backend (`clever features enable system-git`), `
 > [!WARNING]
 > Disabling TLS verification entirely (for example with `NODE_TLS_REJECT_UNAUTHORIZED=0`) exposes you to man-in-the-middle attacks, including the theft of your Clever Cloud credentials. Always prefer trusting your CA with one of the methods above.
 
+## HTTP proxy
+
+On a network where outgoing traffic must go through an HTTP proxy, Clever Tools follows the same `http_proxy` / `https_proxy` environment variables as tools like `curl`, `kubectl` or `s3cmd`. Set them in your shell and every request the CLI makes (API calls, update checks…) is routed through the proxy:
+
+```bash
+# Linux / macOS
+export http_proxy=http://proxy.example.com:3128
+export https_proxy=http://proxy.example.com:3128
+clever profile
+```
+
+```powershell
+# Windows (PowerShell)
+$env:http_proxy = "http://proxy.example.com:3128"
+$env:https_proxy = "http://proxy.example.com:3128"
+clever profile
+```
+
+The uppercase variants (`HTTP_PROXY`, `HTTPS_PROXY`) are also recognized, and a proxy that requires authentication can be configured inline with `http://user:password@proxy.example.com:3128`. To bypass the proxy for some hosts, list them in the `no_proxy` (or `NO_PROXY`) variable:
+
+```bash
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -108,6 +108,33 @@ If you switch to the system git backend (`clever features enable system-git`), `
 > [!WARNING]
 > Disabling TLS verification entirely (for example with `NODE_TLS_REJECT_UNAUTHORIZED=0`) exposes you to man-in-the-middle attacks, including the theft of your Clever Cloud credentials. Always prefer trusting your CA with one of the methods above.
 
+## HTTP proxy
+
+On a network where outgoing traffic must go through an HTTP proxy, Clever Tools follows the same `http_proxy` / `https_proxy` environment variables as tools like `curl`, `kubectl` or `s3cmd`. Set them in your shell and every request the CLI makes (API calls, update checks…) is routed through the proxy:
```
pdesoyres-cc:
```suggestion
On a network where outgoing traffic must go through an HTTP proxy, Clever Tools follows the same `http_proxy` / `https_proxy` environment variables. Set them in your shell and every request the CLI makes (API calls, update checks…) is routed through the proxy:
```

pdesoyres-cc (approved):
LGTM with just a minor nitpick.

Issue #939, opened 2025-07-21
clever login fails if it can't open the browser


BlackYoup:
I'm on a windows machine. It seems like `clever login` can't open the browser for some reason. In that case, the CLI exits with the error "Something went wrong while trying to log you in". Sorry I can't share the logs.

But instead of having an error, we could have the CLI ask us to open the link manually and enter the credentials the API generated?

davlgd:
Can you try [this binary](https://github.com/CleverCloud/clever-tools/pull/940) in your Windows system and tell me if it fixes the "open" issue ? I'll check how to handle login when an error occurs after that

BlackYoup:
> Can you try [this binary](https://github.com/CleverCloud/clever-tools/pull/940) in your Windows system and tell me if it fixes the "open" issue ? I'll check how to handle login when an error occurs after that

Same issue :-/

<img width="2287" height="409" alt="Image" src="https://github.com/user-attachments/assets/431aabef-282a-43b4-b948-22ce0aa1a1d9" />

atyrode:
Confirmed with Clever Tools 4.11.0 on headless Linux through nixpkgs.

Command and complete output (OAuth nonce redacted):

```console
➜  ~ nix shell nixpkgs#clever-tools --command clever login
Opening https://console.clever-cloud.com/cli-oauth?cli_version=4.11.0&cli_token=<redacted> in your browser to log you in…
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: spawn xdg-open ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn xdg-open',
  path: 'xdg-open',
  spawnargs: [
    'https://console.clever-cloud.com/cli-oauth?cli_version=4.11.0&cli_token=<redacted>'
  ]
}

Node.js v22.23.1
```

Because the process exits, it cannot continue polling after the URL is opened manually on another device.

Expected behavior:

- catch browser-launch failures;
- keep polling;
[...]

PR #1111, opened 2026-07-14
fix(login): continue when browser launch fails

atyrode:
Refs #939

## Context

`clever login` opens the Console login URL and then polls the OAuth endpoint. With `open@10.2.0` and `{ wait: false }`, `open()` returns a detached `ChildProcess` without installing an `error` listener.

On a packaged headless Linux/Nix installation without `xdg-open`:

```bash
$ nix shell nixpkgs#clever-tools --command clever login
Opening https://console.clever-cloud.com/cli-oauth?cli_version=4.11.0&cli_token=<redacted> in your browser to log you in…
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: spawn xdg-open ENOENT
```

In bundled Linux builds, if the system `xdg-open` is missing, the child emits an asynchronous `ENOENT` event. That event was unhandled, so Node terminated the CLI before OAuth polling could continue. Catching only the awaited `open()` call does not handle the later child-process event.

## Solution

Handle both browser-launch failure paths at the existing login call site:

- catch synchronous throws and rejected opener promises;
- attach a one-time `error` listener to the returned child process;
- print a static warning asking the user to open the already-displayed URL manually;
- continue the existing OAuth polling flow.

The change remains local to login and preserves `{ wait: false }`. It does not change polling, cancellation, authentication, timeouts, the shared browser helper, launcher exit-code handling, or process lifetime. The launch error is intentionally not logged because its spawn arguments can contain the credential-bearing login URL.

## How to test

- `npm run validate`
- build the Rollup bundle and Linux pkg binary;
- run the packaged binary without `xdg-open` against loopback-only API and Console endpoints;
- verify the process remains alive and continues polling after the launcher error;
- verify a genuine polling failure still exits with code `1`;
- verify Ctrl+C still terminates promptly.

The packaged Linux reproduction passes. The original Windows report remains unverified; please add the `build:win` label so the reporter can test a Windows preview.

## Optional regression coverage

Three dependency-free helper tests are available separately if maintainers want to adopt a test entry point. They are intentionally not part of this source-only PR:

- [test-only diff against this fix](https://github.com/atyrode/clever-tools/compare/fix/login-browser-launch-fallback...optional-tests/login-browser-launch-regression)
- [optional test branch](https://github.com/atyrode/clever-tools/tree/optional-tests/login-browser-launch-regression)
- [standalone test commit](https://github.com/atyrode/clever-tools/commit/a9849f23c529d939c8067c55995794af5bcb155e)


src/commands/login/login.command.js
```diff
@@ -21,6 +21,10 @@ const POLLING_INTERVAL = 2000;
 
 const POLLING_MAX_TRY_COUNT = 60;
 
+function warnManualLogin() {
+  Logger.warn('Could not open your browser automatically. Open the URL above manually to continue logging in.');
+}
+
```
hsablonniere: suggestion: move into `loginViaConsole`, next to its call sites (see the other suggestion).

```suggestion
```

src/commands/login/login.command.js
```diff
@@ -56,7 +60,12 @@ async function loginViaConsole(apiHost, consoleTokenUrl) {
 
   Logger.debug('Try to login to Clever Cloud…');
   Logger.println(`Opening ${styleText('blue', consoleUrl.toString())} in your browser to log you in…`);
-  await open(consoleUrl.toString(), { wait: false });
+  try {
+    const browserProcess = await open(consoleUrl.toString(), { wait: false });
+    browserProcess.once('error', warnManualLogin);
+  } catch {
+    warnManualLogin();
+  }
```
hsablonniere: suggestion: also print `error.message` (`spawn xdg-open ENOENT`) so the user sees why it failed. Safe to show: the message has no URL, only `spawnargs` does.

```suggestion
  function warnBrowserLaunchFailed(error) {
    Logger.warn(`Failed to open your browser automatically (${error.message}).`);
    Logger.println('Open the URL above manually to continue logging in.');
  }

  try {
    const browserProcess = await open(consoleUrl.toString(), { wait: false });
    browserProcess.once('error', warnBrowserLaunchFailed);
  } catch (error) {
    warnBrowserLaunchFailed(error);
  }
```

hsablonniere (changes requested):
Thanks @atyrode, solid fix. I went and checked how a few other CLIs handle this case (gh, gcloud) and left some suggestions inline.

atyrode:
Thanks! I applied both suggestions. GitHub’s batch application introduced mixed line endings in the changed block, so I added a formatting-only commit to normalize them. npm run validate passes.

hsablonniere (approved):
LGTM 👍

hsablonniere:
:pray: Thanks for the fix and for tracking down the async `error` event on the detached child process, that was a subtle one.

Merged manually into `master` at dfa2533d, squashing the three commits into one so the changelog stays readable.

Closing this PR since the change is already on `master`.

PR #1112, opened 2026-08-03
feat(drain): support Splunk HEC as a drain target

KannarFr:
Adds the `splunk` drain type, following the `SPLUNK` recipient merged in OVD3.

```
clever drain create splunk "https://<host>:8088/services/collector/event" --hec-token <token>
clever drain create splunk "https://<host>:8088/services/collector/event" --hec-token <token> \
  --index my-index --sourcetype clevercloud --tls-verification trustful
```

### What's in it

- `SPLUNK` in `DRAIN_TYPES`, which drives the `<drain-type>` enum, its help text and shell completion.
- `--hec-token`, required for this type, sent as `recipient.token`.
- `--index` and `--sourcetype`, both optional and **omitted from the payload when unset**: a field present in the envelope overrides the setting bound to the customer's HEC token, so defaulting them would silently override the server-side configuration.
- `--tls-verification <default|trustful>`, mapped to the API's `TLSMode`. Splunk ships a self-signed certificate on `:8088` by default, so without it a self-hosted HEC is unreachable — for real deliveries *and* for `clever drain check`, since the probe honours the same setting.

### Everything else already works

`clever drain`, `drain get`, `drain check`, `drain enable|disable|remove` are keyed by drain ID and the recipient probe runs server-side, so they needed no change.

One exception worth flagging: `formatDrain()` resolved the type label through an unchecked `DRAIN_TYPES[recipient.type]` lookup, so `clever drain get` on a Splunk drain created from the Console or the API threw a `TypeError`. That's fixed by the registry entry. The same function now also shows `Sourcetype`, and a `TLS verification` line only when verification is relaxed — `DEFAULT` is the implicit norm and printing it on every Elasticsearch drain would be noise.

### Checks

`npm run validate` (lint, prettier, typecheck, docs check) passes. Verified by hand: the help output, the parse-time rejection of an invalid `--tls-verification`, `--index` and `--index-prefix` parsing independently of each other, and shell completion offering `splunk` and the two TLS modes.

Docs under `src/commands/` and `skills/` are regenerated; `docs/services-logs-drains.md` gets a Splunk section and the `--source-token` line that was missing from its option list.

### DIFF
--- a/docs/services-logs-drains.md
+++ b/docs/services-logs-drains.md
@@ -20,6 +20,7 @@ Where `DRAIN-TYPE` is one of:
 - `newrelic`: for NewRelic endpoint (note that this endpoint needs your NewRelic API Key)
 - `ovh-tcp`: for OVH TCP syslog endpoint (note that this endpoint has an optional sd-params parameter)
 - `raw-http`: for HTTP endpoint (note that this endpoint has optional username/password parameters as HTTP Basic Authentication)
+- `splunk`: for Splunk HTTP Event Collector endpoint (note that this endpoint needs an HEC token)
 - `syslog-tcp`: for TCP syslog endpoint
 - `syslog-udp`: for UDP syslog endpoint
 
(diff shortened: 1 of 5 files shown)


hsablonniere:
Small code tweak: Splunk options are now destructured together with the other drain-type-specific options, instead of a separate line.

hsablonniere (approved):
LGTM :+1:

PR #1114, opened 2026-08-03
feat(cli): group options by the context they apply to in help and docs

KannarFr:
An option definition can now declare the contexts it applies to, and help, reference docs and the LLM documentation render one section per context instead of a single flat list.

```js
tlsVerification: defineOption({
  name: 'tls-verification',
  schema: z.enum(['default', 'trustful']).optional(),
  description: 'TLS verification mode, use `trustful` to accept a self-signed certificate',
  placeholder: 'tls-verification',
  groups: ['splunk'],
}),
```

### Why

`clever drain create` carries eleven options, and only three of them apply to every drain. The other eight are specific to one or two drain types, and the only thing saying so was a `(for elasticsearch)` suffix hand-written at the end of each description. Reading the help meant scanning a flat list and filtering it by eye, and each new drain type makes it worse.

Those suffixes were also unreliable, because nothing tied them to the handler. `--sd-params` has been documented as `(for ovh-tcp)` while the handler has always accepted it for `syslog-tcp` and `syslog-udp` too — this PR fixes that as a side effect of declaring the groups.

`clever drain create --help` now reads:

```
OPTIONS
      --addon <addon-id>                       Add-on ID or real ID
  -a, --alias <alias>                          Short name for the application
      --app <app-id|app-name>                  Application to manage by its ID (or name, if unambiguous)

OPTIONS FOR BETTERSTACK
  -t, --source-token <source-token>            Source token

OPTIONS FOR ELASTICSEARCH
  -i, --index-prefix <index-prefix>            Optional index prefix, `logstash` value is used if not set
  -p, --password <password>                    Basic auth password
  -u, --username <username>                    Basic auth username
...
OPTIONS FOR SPLUNK
      --hec-token <hec-token>                  HTTP Event Collector token
      --index <index>                          Optional target index, the HEC token's own index is used if not set
      --sourcetype <sourcetype>                Optional sourcetype, the HEC token's own sourcetype is used if not set
      --tls-verification <tls-verification>    TLS verification mode, use `trustful` to accept a self-signed certificate (default, trustful)
```

### Design notes

- **An option can declare several groups and is then repeated in each.** `--username` shows up under both `elasticsearch` and `raw-http`, so a section is the complete answer to "what can I pass with this type?" rather than a remainder you have to combine with the ungrouped list. It does make the help longer overall — eight headings for fourteen rows where there were ten flat rows — which is the deliberate trade for not having to filter it by eye.
- **The split is computed once**, in `getCommandInfo` as `optionGroups`, so terminal help, the `src/commands/**` reference docs and the LLM documentation stay consistent by construction. `options` is unchanged and still lists every option exactly once for any other consumer.
- **Ungrouped stays the default.** A command whose options declare no group renders exactly as before. The check on that: regenerating all 51 doc files touches only the two drain ones.
- **Grouping is presentational, parsing is untouched.** Passing a Splunk option to a `syslog-tcp` drain is accepted and ignored today, and still is. Making it an error would mean turning each drain type into its own subcommand, which is a behaviour change for anything that passes flags generically — worth considering separately, and this metadata is what it would build on.

### Checks

`npm run validate` passes. Spot-checked `clever deploy --help` (ungrouped, unchanged) and `clever version --help` (no options at all).

Based on #1112 rather than `master`, since the Splunk options are what motivated this; it retargets to `master` automatically once #1112 merges.


KannarFr:
Do you have a better strategy for this?

hsablonniere:
@KannarFr Sorry this was automatically closed because you created a PR based on another branch and I cannot reopen it because the branch was deleted.

We discussed it in the team meeting this morning. I understand the problem, I like your suggestion but I'm going to try the multiple sub commands.

PR #1116, opened 2026-08-03
chore: remove unused @babel/traverse dependency

### LINKED ISSUE #1115: @babel/traverse is declared but never imported (16 packages in every user install)
## Summary

`@babel/traverse` is listed in `dependencies` but is never imported. It reaches every user
who installs `clever-tools` and adds 16 packages to their tree.

## Where it came from

It was added by `3949a5b` (2025-12-17), *"refactor: apply automated script to generate new
command structure"*. That commit's only change to `package.json` was this one line:

```
+    "@babel/traverse": "7.28.5",
```

The codemod that needed it was a one-off — `commands2/` is not in the repository and no
script under `scripts/` references babel. The dependency stayed behind, in `dependencies`
rather than `devDependencies`.

## Verification

At `707580b`, Node 22.14.0:

- `git grep -i traverse` matches `package.json:52` and nothing else. Grepping the whole
  working tree for `babel` outside `package-lock.json` also matches only `package.json`.
- Nothing else in the lockfile depends on `@babel/traverse` — the root is its only
  requester.
- `npm run validate` (lint, format:check, typecheck, docs:check) before removing it:
  exit 0, "All docs up-to-date (51 files)". After: identical, exit 0.
- `node bin/clever.js --version` still prints `4.11.0`.

## Two different numbers, both measured

These are worth separating, because they answer different questions:

| | entries removed |
| --- | --- |
| **What a user who installs `clever-tools` loses** | **16** (195 → 179) |
| What this repository's `node_modules` loses | 3 |

The repo only drops `@babel/traverse`, `@babel/template` and `@babel/helper-globals`,
because `@yao-pkg/pkg` keeps `@babel/generator`, `@babel/parser` and `@babel/types` for
building binaries. `@yao-pkg/pkg` is a devDependency, so an end user never installs it and
the whole chain leaves for them: `@babel/code-frame`, `generator`, `helper-globals`,
`helper-string-parser`, `helper-validator-identifier`, `parser`, `template`, `traverse`,
`types`, the four `@jridgewell/*`, `js-tokens`, `jsesc` and `picocolors`.

Both figures come from resolving two real lockfiles and diffing them, not from reading the
dependency graph on paper.

## Separately, and I am *not* proposing this

`eventsource` is also declared at the root and never imported. Removing it changes nothing:
195 entries either way, because `@clevercloud/client` depends on it and keeps it in the
tree. Since every runtime dependency here is pinned to an exact version, I assume the root
entry may be deliberate — pinning the version that gets hoisted. I have left it alone. If
it is not deliberate, it is one more line that can go.

## Note on a possible pitfall

If anyone reproduces this on Windows, clone with `core.autocrlf=false`. Otherwise
`npm run validate` fails `format:check` on 333 files with line-ending noise and hides the
real result. That is a local artifact, not a problem with the repository.

---

I have the change ready and will open a PR referencing this issue, per CONTRIBUTING asking
that an issue exist first. Happy to close this instead if you would rather keep the
dependency for a reason I cannot see from outside.

HeversonSilva-gif:
Closes #1115.

`@babel/traverse` is declared in `dependencies` and never imported. It was added by
`3949a5b` ("refactor: apply automated script to generate new command structure",
2025-12-17), whose only `package.json` change was that one line. The codemod was a one-off
— `commands2/` is not in the repository and nothing under `scripts/` mentions babel — and
the entry stayed behind in `dependencies` rather than `devDependencies`.

**Measured, at `707580b` on Node 22.14.0:**

| | |
| --- | --- |
| `npm run validate` before | exit 0, "All docs up-to-date (51 files)" |
| `npm run validate` after | exit 0, identical |
| `node bin/clever.js --version` after | `4.11.0` |
| Entries a **user** stops installing | **16** (195 → 179) |
| Entries this repo's `node_modules` loses | 3 |

The two dependency figures differ because `@yao-pkg/pkg` keeps `@babel/generator`,
`@babel/parser` and `@babel/types` for building binaries. It is a devDependency, so an end
user never installs it and the whole chain leaves for them. Both numbers come from
resolving two real lockfiles and diffing them.

`grep -i traverse` across the working tree matches `package.json:52` and nothing else, and
nothing else in the lockfile requires `@babel/traverse` — the root is its only requester.

I left `eventsource` alone even though it is also declared and never imported here, because
removing it changes nothing (`@clevercloud/client` keeps it in the tree) and the exact pin
may be deliberate. That is noted in the issue rather than acted on.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "4.11.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@babel/traverse": "7.28.5",
         "@clevercloud/client": "12.0.0",
         "@inquirer/prompts": "7.8.4",
         "char-regex": "2.0.2",
@@ -1001,6 +1000,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.28.6.tgz",
       "integrity": "sha512-JYgintcMjRiCvS8mMECzaEn+m3PfoQiyqukOMCCVQtoJGYJw8j/8LBJEiqkHLkfwCcs74E3pbAUFNg7d9VNJ+Q==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-validator-identifier": "^7.28.5",
@@ -1015,6 +1015,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.28.6.tgz",
       "integrity": "sha512-lOoVRwADj8hjf7al89tvQ2a1lf53Z+7tiXMgpZJL3maQPDxh0DgLMN62B2MKUOFcoodBHLMbDM6WAbKgNy5Suw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/parser": "^7.28.6",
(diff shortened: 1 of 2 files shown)


hsablonniere:
Thanks for the thorough investigation, this was clearly an oversight: `@babel/traverse` was only ever needed by the one-off codemod that generated the new command structure (`3949a5b`), and nothing in the repo has referenced it since. Good catch on the 16-package install-time saving.

Confirmed `eventsource` is a different case and should stay: it's a peerDependency of `@clevercloud/client` (`^4.0.0`), not a transitive dependency, so it has to be declared at the root here for that package to work. Right call leaving it alone.

hsablonniere (approved):
LGTM :+1:

PR #1121, opened 2026-08-12
fix: use update-notifier message template for non-npm installs

### LINKED ISSUE #1117: update-notifier: getDetails() is never called and tagsUrl is ignored, so non-npm users are told to run npm i -g
## Summary

`src/initial-update-notifier.js` passes two options to `update-notifier` that it does not
support and silently ignores. One of them is the reason the file exists.

```js
updateNotifierModule({
  pkg,
  tagsUrl: 'https://api.github.com/repos/CleverCloud/clever-tools/tags',   // ignored
}).notify({
  isGlobal: true,
  getDetails() {                                                          // never called
    const docsUrl = '…/docs#how-to-use-clever-tools';
    return `\nPlease follow this link to update your clever-tools:\n${docsUrl}`;
  },
});
```

## Why it matters

`clever-tools` ships through Docker images, binary downloads, RPM and DEB packages,
Homebrew, WinGet and AUR, as the README says. `getDetails` looks written to keep those users
from being told to reinstall through npm.

It never runs, so they are told exactly that. This is the box a user actually gets:

```
   ╭─────────────────────────────────────────╮
   │                                         │
   │    Update available 4.11.0 → 9.9.9      │
   │   Run npm i -g clever-tools to update   │
   │                                         │
   ╰─────────────────────────────────────────╯
```

Someone who installed via `brew` or `apt` is being pointed at a command that will not
update their installation.

## Verification

Neither option exists in `update-notifier`. Grepping the whole published package for
`tagsUrl` and `getDetails` returns nothing, and that holds for **5.1.0, 6.0.2, 7.0.0 and
7.3.1** — `getDetails` has never been part of its API, so this is not a regression from a
version bump.

The options `notify()` reads are `defer`, `message`, `isGlobal` and `boxenOptions`; the
constructor reads `pkg`, `packageName`, `packageVersion`, `updateCheckInterval`,
`shouldNotifyInNpmScript` and `distTag`.

Then measured rather than inferred — your exact call, with a seeded cache and an isolated
`XDG_CONFIG_HOME`, on Node 22.14.0:

```
getDetails FOI CHAMADO?      NAO
texto dele aparece na saida? NAO
bytes renderizados:          611
```

The box above is that run's real output.

**`tagsUrl` is the same story.** `update-notifier` only ever queries the npm registry for
the latest version; there is no code path that reads a tags URL. So the version being
compared is the npm one, not your GitHub tags — which may or may not be what you want, but
it is not what the option says.

## Possible fix

`notify()` accepts a `message` template with `{currentVersion}`, `{latestVersion}` and
`{packageName}` placeholders. The details text could go there directly, which is one line
and needs nothing new:

```js
.notify({
  isGlobal: true,
  message: `Update available {currentVersion} → {latestVersion}\nPlease follow this link to update your clever-tools:\n${docsUrl}`,
})
```

I have not sent that as a PR because the wording is yours and I did not want to invent it.
Happy to open one if you tell me what it should say. If `tagsUrl` was meant to make the
check use GitHub tags rather than npm, that is a larger change and I would not guess at it.

Environment: Node 22.14.0, npm 10.9.2, `clever-tools` at `707580b`, `update-notifier@7.3.1`.

mehdi653:
## Summary
update-notifier ignores getDetails and tagsUrl. This uses the supported message template and points to the clever-tools docs update section so brew/apt/binary users are not told to run a global npm install.

Fixes #1117

## Test plan
- Trigger the update notice and confirm the docs link appears
- Confirm the notice no longer recommends a global npm install

### DIFF
--- a/src/initial-update-notifier.js
+++ b/src/initial-update-notifier.js
@@ -5,14 +5,13 @@ import { hasParam } from './lib/has-param.js';
 // These need to be set before Logger and other stuffs
 const updateNotifierExplicitFalse = hasParam('--no-update-notifier') || hasParam('--update-notifier', 'false');
 if (!updateNotifierExplicitFalse) {
+  const docsUrl = 'https://github.com/CleverCloud/clever-tools/tree/master/docs#how-to-use-clever-tools';
+  // update-notifier ignores getDetails/tagsUrl. Use the supported message template instead
+  // so brew/apt/binary users are not told to run npm i -g.
   updateNotifierModule({
     pkg,
-    tagsUrl: 'https://api.github.com/repos/CleverCloud/clever-tools/tags',
   }).notify({
     isGlobal: true,
-    getDetails() {
-      const docsUrl = 'https://github.com/CleverCloud/clever-tools/tree/master/docs#how-to-use-clever-tools';
-      return `\nPlease follow this link to update your clever-tools:\n${docsUrl}`;
-    },
+    message: `Update available {currentVersion} -> {latestVersion}\nPlease follow this link to update your clever-tools:\n${docsUrl}`,
   });
 }


hsablonniere:
:pray: Thanks @mehdi653 for your work, I'll follow up in #1133 with a dedicated docs page. I just need to check that my colleagues agree on that.

PR #1123, opened 2026-08-12
feat(ssh): add --instance option to target a running instance by ID

mehdi653:
## Summary
When several instances are running, clever ssh currently requires an interactive prompt and fails in non-interactive mode. This adds --instance instance-id so users and scripts can target a specific running instance without a prompt.

Instance IDs are used rather than instance numbers, which are not stable across deployments (per hsablonniere feedback on #1067). Support for number / any / all selection is left for a follow-up (judu).

Fixes #1067

## Change
- Add --instance instance-id option to clever ssh
- When set, select that running instance and skip the interactive prompt
- When the ID does not match a running instance of the app, fail with a clear error
- Update ssh docs and generated skill reference

## Spec notes
- Selection key is the Clever Cloud instance ID (not the ephemeral instance number)
- Existing behavior is unchanged when --instance is omitted (single instance auto-select, multi-instance TTY prompt, non-TTY failure)
- Works with -c / --command for non-interactive one-shot commands

## Test plan
- [x] run validate on this branch
- [x] valid instance id connects without prompt
- [x] bogus instance id fails with a clear error
- [x] multi-instance non-TTY with option succeeds
- [x] multi-instance non-TTY without option still fails as before

## E2E results (2026-08-13)

Tested against a real Clever Cloud Node app with the local CLI from this PR checkout (HOME/.local/bin clever, which exposes --instance).

App: ssh-instance-e2e-1123 (app_99596987-df36-4925-8bab-26f50b2e1ba5), scaled to 2 instances, dedicated ed25519 key registered on the account.

1. Valid instance id, agent disabled, remote echo:
   -> OK, exit 0, no prompt
2. Bogus instance id:
   -> clear error: Instance bogus is not a running instance of this application, exit 1
3. Non-TTY multi-instance WITH --instance (stdin closed) on the second instance id:
   -> OK, exit 0
4. Non-TTY multi-instance WITHOUT --instance:
   -> clear error: Multiple instances are running. Cannot select in non-interactive mode., exit 1

Also: validate passed on 5898785 (lint, prettier, typecheck, docs:check). Working tree clean; no docs rewrite and no force-push needed.

Test app and temporary SSH key were deleted after the run.


mehdiLaMine:
## E2E results (2026-08-13)

Tested against a real Clever Cloud Node app with the local CLI from this PR checkout (HOME/.local/bin clever, which exposes --instance).

App: ssh-instance-e2e-1123 (app_99596987-df36-4925-8bab-26f50b2e1ba5), scaled to 2 instances, dedicated ed25519 key registered on the account.

1. Valid instance id, agent disabled, remote echo:
   -> OK, exit 0, no prompt
2. Bogus instance id:
   -> clear error: Instance bogus is not a running instance of this application, exit 1
3. Non-TTY multi-instance WITH --instance (stdin closed) on the second instance id:
   -> OK, exit 0
4. Non-TTY multi-instance WITHOUT --instance:
   -> clear error: Multiple instances are running. Cannot select in non-interactive mode., exit 1

Also: validate passed on 5898785 (lint, prettier, typecheck, docs:check). Working tree clean; no docs rewrite and no force-push needed.

Test app and temporary SSH key were deleted after the run.

Note: PR body PATCH still blocked from available tokens (MCP mehdi653 fine-grained PAT 403 on CleverCloud/clever-tools; gh CLI mehdiLaMine lacks UpdatePullRequest). Please paste the proposed Summary/Change/Spec/Test plan/E2E sections from the earlier comment into the PR description, and mark the validated/E2E checkboxes as done.

src/commands/ssh/ssh.command.js
```diff
@@ -28,12 +28,18 @@ export const sshCommand = defineCommand({
       aliases: ['c'],
       placeholder: 'command',
     }),
+    instance: defineOption({
+      name: 'instance',
+      schema: z.string().optional(),
```
davlgd: `z.string()` accepts the empty string, so `--instance ""` gets past the `instance != null` check below and reports `Instance  is not a running instance of this application` — with a gap where the ID should be.

```suggestion
      schema: z.string().min(1).optional(),
```

Checked against the bundled Zod: `""` is accepted today and rejected with `.min(1)`. It doesn't trim, so `" "` still gets through, but that is a separate question.

davlgd: `z.string()` accepts the empty string, so a missing value reaches the handler and reports `Instance  is not a running instance of this application`, with a gap where the ID should be. It happens with `--instance` alone, and with `--instance --command "hostname"`, where the flag takes no value and the command still runs.

```suggestion
      schema: z.string().min(1).optional(),
```

Both are then rejected at parse time instead. `if (instance != null)` stays as it is: cliparse passes `null` when the option is absent, and the loose check already covers that.

davlgd (approved):
Thanks for this! Just a nitpick here

src/commands/ssh/ssh.command.js
mehdi653: Good catch, thanks! Pushed `z.string().min(1).optional()` so an empty `--instance` fails at parse time.

hsablonniere:
Thanks a lot @mehdi653 for this contribution!

To keep the history clean before merging, we squashed your commits into a single one and reworked the commit message wording (you're still the author). We also replaced `Fixes #1067` with `Refs #1067` in the description, so the issue stays open: we'll handle the rest (selection by number, `any` / `all`) in another PR.

hsablonniere (approved):
LGTM 👍

Issue #996, opened 2025-11-13
Adapt `clever-tools` update command based on the package manager


maximepvrt:
Currently, when a new version of `clever-tools` is available, the message shown is:

```
Update available 3.13.1 → 4.4.0
Run npm i -g clever-tools to update
```

This assumes the user is using **npm**, while some users may be using **pnpm, yarn, brew**, etc.

**Proposal:**

* Detect the package manager in use or display multiple commands for different environments:

  * npm: `npm i -g clever-tools`
  * pnpm: `pnpm add -g clever-tools`
  * yarn: `yarn global add clever-tools`
  * brew: `brew upgrade clever-tools`

* Dynamically show the appropriate update command based on the user’s environment.

**Expected benefits:**

* Reduce update errors.
* Provide a smoother experience for all users.
* Ensure consistent support for different package managers.

hsablonniere:
Hey @maximepvrt,

Following up on this: we shipped #1133, which takes a different route than detecting your package manager.

Instead of trying to guess whether you installed via npm, pnpm, Yarn or Bun (a game we didn't want to play, per the comment above), we went the other way: every install channel now shows the update notice, including the ones that never had it before (Homebrew, WinGet, apt/yum, AUR, Exherbo, the standalone binary). And rather than print a one-size-fits-all command, the message now points to a dedicated guide where you pick the right one for however you installed: https://github.com/CleverCloud/clever-tools/blob/master/docs/update.md

Simpler for us to maintain, and covers every channel instead of just the npm family. Closing this one, feel free to reopen if it misses your case.

PR #1125, opened 2026-08-19
feat(aur): publish a Node.js flavored clever-tools package

### LINKED ISSUE #1124: Publish the clever-tools AUR package from our CI
We now own the [`clever-tools`](https://aur.archlinux.org/packages/clever-tools) AUR package, previously maintained by a third party. Our CI only publishes [`clever-tools-bin`](https://aur.archlinux.org/packages/clever-tools-bin) today.

Unlike `clever-tools-bin`, this package is built from sources, so it should ship the Node.js version of the CLI rather than the prebuilt binary.

To investigate:

- confirm our AUR account (the SSH key used by the release workflow) is listed as maintainer of `clever-tools`
- review the inherited PKGBUILD and decide what the package should actually build (Node.js bundle running on the system `nodejs`, instead of rebuilding the same `pkg` binary as `clever-tools-bin`)
- extend `scripts/publish-aur.js` and the release workflow to publish both packages
- document both packages in `docs/setup-systems.md`

hsablonniere:
Closes #1124

## Context

We now own the [`clever-tools`](https://aur.archlinux.org/packages/clever-tools) AUR package. It used to be maintained by a third party, and its inherited PKGBUILD rebuilt, from the GitHub sources, the very same `@yao-pkg/pkg` binary that we already publish as [`clever-tools-bin`](https://aur.archlinux.org/packages/clever-tools-bin): two AUR packages, one artifact, and a bump lag of days to weeks after each release.

It now ships the Node.js flavor of the CLI, running on the system `nodejs` package, and both packages are published by the release pipeline.

What Arch users get:

- `clever-tools-bin`: self-contained binary, no runtime dependency, frozen Node.js runtime
- `clever-tools`: Node.js flavor, `arch=any`, follows the system `nodejs`, so runtime security fixes are applied by pacman

## Proposal

`scripts/publish-aur.js` now takes a flavor argument (`bin` or `nodejs`) and templates live in `scripts/templates/aur/<flavor>/`. One script, two AUR repositories, two release jobs: `publish-aur-bin` (unchanged, `needs: publish-cellar-archives`) and `publish-aur-nodejs` (`needs: publish-npm`).

### Design decisions

**The npm registry tarball as source.** This is the URL template prescribed by the [Arch Node.js package guidelines](https://wiki.archlinux.org/title/Node.js_package_guidelines): immutable, versioned, and already attested by our `npm publish --provenance`. It is also what most Node.js CLIs packaged on the AUR use (`gemini-cli`, `netlify`, `firebase-tools`, `devcontainer-cli`, `repomix`).

**`npm ci` against the lockfile, not the wiki's `npm install -g --prefix`.** This is the one place where we knowingly deviate from the guidelines, and the reason is worth stating.

npm always strips `package-lock.json` from published tarballs, even when the file is explicitly listed in the `files` field, so the minimal recipe resolves the transitive dependency tree to whatever is latest at build time. Measured on 4.11.0: **52 transitive dependencies** resolve to versions this release was never tested against, including major bumps (`@pnpm/npm-conf` 2.3.1 → 3.0.3, `ansi-regex` 5.0.1 → 6.3.0, `@inquirer/core` 10.2.0 → 10.3.2). Our direct dependencies are pinned to exact versions, their own dependencies are not.

Three consequences we are not willing to ship:

- users would not install what we test and release: a regression introduced in a transitive dependency the week after a release lands on Arch users without any version of `clever-tools` moving
- the lockfile carries an integrity hash per package; without it nothing is verified, and a compromised transitive dependency published after our release goes straight into the package
- two people building the same `pkgver-pkgrel` two weeks apart get different packages

So the PKGBUILD fetches `package-lock.json` from the matching git tag as a second source, and runs `npm ci --omit=dev --ignore-scripts`. The rationale is also inlined in the PKGBUILD, since that file is read by AUR users and by whoever edits it next.

**`--ignore-scripts`**: no production dependency needs a lifecycle script, so none runs while packaging.

**`arch=any`**: verified, no native addon in the production tree.

**A thin `sh` wrapper for `/usr/bin/clever`**, exporting `NO_UPDATE_NOTIFIER=1`: a pacman-managed CLI must not advertise out-of-band updates. Same approach as `command-code`, `claude-code` and `sillytavern` on the AUR.

**No `jq` cleanup of the `_where` and `man` keys**: the recipe in the wiki dates back to npm 6, npm 11 no longer writes those references (verified on the built package).

**`.SRCINFO` stays templated**, and no containerized validation job was added. `makepkg --printsrcinfo` output is byte-identical to the templated file for both flavors, and what a build in CI would catch (a broken PKGBUILD) only happens when the template is edited, not on every tag. The local validation procedure is documented in `CONTRIBUTING.md` instead.

### Drive-by fix

`namcap` reports **3 errors** on the current `clever-tools-bin`: the PKGBUILD declares no dependency at all, while the binary is dynamically linked against glibc, libstdc++ and libgcc. Fixed with `depends=('glibc' 'gcc-libs')`, plus `!debug` to stop shipping an empty `/usr/src/debug`. Both packages now build with zero namcap error.

## How to review

- `scripts/templates/aur/nodejs/PKGBUILD` first, it carries the whole reasoning
- then `scripts/publish-aur.js` for the flavor dispatch, and `scripts/lib/npm-registry.js` which derives the tarball sha512 from `dist.integrity` without downloading anything
- `.github/workflows/release.yml` last, the new job is a twin of the existing one

## How to test

Requires an Arch machine, `makepkg` and `namcap`. Against a local bare repository, nothing reaches AUR:

```
git init --bare -b master /tmp/aur-clever-tools.git
git clone /tmp/aur-clever-tools.git /tmp/aur-seed
git -C /tmp/aur-seed commit --allow-empty -m init
git -C /tmp/aur-seed push origin HEAD:master

AUR_GIT_URL=/tmp/aur-clever-tools.git ./scripts/publish-aur.js 4.11.0 nodejs

mkdir -p /tmp/aur-build && cp git-aur/PKGBUILD /tmp/aur-build/ && cd /tmp/aur-build
makepkg -f
makepkg --printsrcinfo | diff - ~/clever-tools/git-aur/.SRCINFO
namcap PKGBUILD && namcap clever-tools-*.pkg.tar.zst
sudo pacman -U ./clever-tools-4.11.0-1-any.pkg.tar.zst
clever version
```

Results on 4.11.0: package builds as `any`, 5.4 MB compressed / 31 MB installed, zero namcap error, `clever version` and `clever help` work, bash and zsh completions installed, and installing it correctly prompts to remove `clever-tools-bin` thanks to the new `conflicts`.

Same sequence with the `bin` flavor to check the existing channel still builds.

## Out of scope

- no CI job validating PKGBUILD templates; if we want one later, the right trigger is a PR touching `scripts/templates/aur/**`, not every tag

## Before merging

The `AUR_NODEJS_GIT_URL` repository variable is already created (`ssh://aur@aur.archlinux.org/clever-tools.git`). Remaining check: confirm the CI SSH key has push rights on the transferred package, with `ssh aur@aur.archlinux.org list-repos`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -132,8 +132,8 @@ jobs:
           CC_CLEVER_TOOLS_RELEASES_CELLAR_KEY_ID: ${{ secrets.CC_CLEVER_TOOLS_RELEASES_CELLAR_KEY_ID }}
           CC_CLEVER_TOOLS_RELEASES_CELLAR_SECRET_KEY: ${{ secrets.CC_CLEVER_TOOLS_RELEASES_CELLAR_SECRET_KEY }}
 
-  publish-aur:
-    name: 'Publish to AUR'
+  publish-aur-bin:
+    name: 'Publish to AUR (binary package)'
     needs: publish-cellar-archives
     runs-on: ubuntu-latest
     steps:
@@ -150,9 +150,29 @@ jobs:
       - uses: ./.github/actions/download-artifacts
         with: { pattern: build-linux-archive }
       - name: 'Publish to AUR'
-        run: ./scripts/publish-aur.js ${{ github.ref_name }}
+        run: ./scripts/publish-aur.js ${{ github.ref_name }} bin
         env:
-          AUR_GIT_URL: ${{ vars.AUR_GIT_URL }}
+          AUR_GIT_URL: ${{ vars.AUR_BIN_GIT_URL }}
+
+  publish-aur-nodejs:
+    name: 'Publish to AUR (Node.js package)'
+    needs: publish-npm
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Setup SSH key'
+        uses: webfactory/ssh-agent@v0.9.1
+        with:
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+      - name: 'Add AUR to known hosts'
+        run: |
+          mkdir -p ~/.ssh
+          ssh-keyscan -H aur.archlinux.org >> ~/.ssh/known_hosts
+      - uses: actions/checkout@v5
+      - uses: ./.github/actions/setup-node
+      - name: 'Publish to AUR'
+        run: ./scripts/publish-aur.js ${{ github.ref_name }} nodejs
+        env:
+          AUR_GIT_URL: ${{ vars.AUR_NODEJS_GIT_URL }}
 
   publish-homebrew:
     name: 'Publish to Homebrew'
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,15 @@ We provide several utility scripts to streamline development (located in `/scrip
   node scripts/check-github-actions.js
   ```
 
+#### AUR Packages
+
+- **`scripts/publish-aur.js`**: Publish a package flavor to the Arch User Repository
+
+  Both flavors are published from templates in `scripts/templates/aur/<flavor>/`. The `.SRCINFO` is
+  templated alongside the `PKGBUILD`, so both files must be edited together and kept in sync
+  (`makepkg --printsrcinfo` regenerates it for comparison). On an Arch Linux machine, a template
+  change can be validated end to end against a local bare repository, without ever touching AUR.
+
 #### Preview Management
 
 - **`scripts/preview.js`**: Build and manage PR preview versions
@@ -123,6 +132,7 @@ clever-tools/
 ├── scripts/                             # Build and CI scripts (TypeScript with JSDoc enforced)
 │   ├── lib/                             # Shared utilities for scripts
 │   └── templates/                       # Templates for package managers (AUR, Homebrew, etc.)
+│       └── aur/{bin,nodejs}/            # One directory per AUR package flavor
 └── ...
 ```
 
@@ -694,7 +704,7 @@ Our CI/CD pipeline ensures quality and automates releases:
 - Conventional commits drive changelog generation
 - Release-please manages versioning
 - Automated publishing to multiple platforms:
-  - **AUR**: Arch Linux User Repository
+  - **AUR**: Arch Linux User Repository, two packages (`clever-tools`, the Node.js flavor, and `clever-tools-bin`, the self-contained binary)
   - **Cellar**: Clever Cloud's object storage for direct downloads
   - **Docker Hub**: Official Docker images
   - **Exherbo**: Linux distribution packages
@@ -742,6 +752,9 @@ graph LR
         BUNDLE2 --> BUILD_WIN2[build-windows]
     end
 
+    REL --> NPM[publish-npm]
+    NPM --> AUR_NODEJS[publish-aur-nodejs]
+
     BUILD_2 --> PACKAGE_RPM[package-rpm]
     BUILD_2 --> PACKAGE_DEB[package-deb]
     BUILD_2 --> CELLAR_ARC[publish-cellar-archives]
@@ -756,7 +769,7 @@ graph LR
     PACKAGE_DEB --> NEXUS_DEB[publish-nexus-deb]
 
     %% Package managers (depend on cellar archives)
-    CELLAR_ARC --> AUR[publish-aur]
+    CELLAR_ARC --> AUR_BIN[publish-aur-bin]
     CELLAR_ARC --> HOMEBREW[publish-homebrew]
     CELLAR_ARC --> DOCKER[publish-dockerhub]
     CELLAR_ARC --> EXHERBO[publish-exherbo]
--- a/docs/setup-systems.md
+++ b/docs/setup-systems.md
@@ -54,12 +54,21 @@ yarn global add clever-tools
 
 ### Arch Linux (AUR)
 
-If you use Arch Linux, install packages [from AUR](https://aur.archlinux.org/packages/clever-tools-bin/). If you don't know how to use this, run:
+If you use Arch Linux, install Clever Tools from AUR. Two packages are available, they provide the same `clever` command and cannot be installed side by side:
+
+- [`clever-tools`](https://aur.archlinux.org/packages/clever-tools/): the Node.js flavor, it runs on the `nodejs` package of your system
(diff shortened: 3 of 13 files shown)


hsablonniere:
### Follow-up: how do we disable the update notifier properly?

This PR ships a `sh` wrapper exporting `NO_UPDATE_NOTIFIER=1`. It works, but it papers over the fact that we have no real notion of "this install is managed by a package manager".

How it works today, for the record: nothing at the packaging level disables the notifier. `rollup.config.js` strips the import from the entry point when bundling:

```js
if (id.endsWith('/bin/clever.js')) {
  return code.replace("import '../src/initial-update-notifier.js';", '');
}
```

The module and the `update-notifier` dependency are then tree-shaken. Verified on the generated bundle: only the definition of the `--update-notifier` CLI option remains, no `updateNotifierModule`, no `tagsUrl`. Since every binary comes from that bundle, `.deb`, `.rpm`, archives, Homebrew, Docker Hub and `clever-tools-bin` all inherit that behaviour for free. `npm i -g clever-tools` is the only distribution where the notifier actually runs.

Which means the disabling is a side effect of the binary build, not a decision. `clever-tools` (Node.js flavor) is our first distro package that runs the npm sources instead of a binary, so it is the first one where the question has to be answered explicitly, hence the wrapper.

Two things worth fixing beyond this PR:

- `--update-notifier` / `--no-update-notifier` is still advertised by every binary, where it has been a no-op for a while. Same family as #1117.
- we should know how we were installed, and act on it. Rather than silencing the notifier, a package-managed install should say `pacman -Syu clever-tools`, `apt upgrade clever-tools`, `brew upgrade clever-tools`, and so on. That is exactly #996.

Options for the "how do we know" part, roughly by decreasing cleanliness:

1. an installation marker written by each package (a field added to the installed `package.json`, or a small file next to it), read at runtime. Explicit, works for every channel, and gives #996 the information it needs.
2. entry point path detection (`/usr/lib`, `/usr/share`, `/opt`, the Homebrew Cellar, ...). No packaging change, but heuristic.
3. an environment variable set by a wrapper, what this PR does. Simple, per package, and stays useful as a fallback whatever we pick.

No decision needed to merge this PR, the wrapper is not a blocker. Opening the discussion so we do not end up with one ad hoc mechanism per packaging channel.

docs/setup-systems.md
```diff
@@ -21,14 +21,23 @@ Clever Cloud CLI is based on Node.js. We thought it to be easily available on an
...
+- [`clever-tools-bin`](https://aur.archlinux.org/packages/clever-tools-bin/): a self-contained binary, with no runtime dependency
+
+If you don't know how to use this, run:
 
 ```
-git clone https://aur.archlinux.org/clever-tools-bin.git clever-tools
+git clone https://aur.archlinux.org/clever-tools.git
 cd clever-tools
 makepkg -si
 ```
```
pdesoyres-cc, line 66: I'm not sure AUR consumers often use this way of installing packages.

fix: replace that by a `yay` command instead of a manual `git clone` and `makepkg`

And add the alternative command for installing the `clever-tools-bin` package. and remove the TIL below

CONTRIBUTING.md
```diff
@@ -86,6 +86,33 @@ We provide several utility scripts to streamline development (located in `/scrip
...
+  cp git-aur/PKGBUILD /tmp/aur-build/ && cd /tmp/aur-build
+  makepkg -f
+  makepkg --printsrcinfo | diff - ../../git-aur/.SRCINFO   # template must match
+  namcap PKGBUILD && namcap clever-tools-*.pkg.tar.zst
+  sudo pacman -U ./clever-tools-*.pkg.tar.zst && clever version
+  ```
+
+  The `.SRCINFO` is templated alongside the `PKGBUILD`, so both files must be edited together: the
+  `makepkg --printsrcinfo` diff above is what catches a desync.
+
```
pdesoyres-cc, line 97: remove: Not sure this level of details is necessary here

.github/workflows/release.yml
```diff
@@ -150,10 +150,30 @@ jobs:
       - uses: ./.github/actions/download-artifacts
         with: { pattern: build-linux-archive }
       - name: 'Publish to AUR'
-        run: ./scripts/publish-aur.js ${{ github.ref_name }}
+        run: ./scripts/publish-aur.js ${{ github.ref_name }} bin
         env:
           AUR_GIT_URL: ${{ vars.AUR_GIT_URL }}
```
pdesoyres-cc: question: should we rename this env var to `AUR_BIN_GIT_URL` (in both vars injection from GitHub variables, and env var injected to our build script)

scripts/templates/aur/nodejs/PKGBUILD
```diff
@@ -0,0 +1,68 @@
+# Maintainer: <%= maintainer %>
+# Contributor: doclic <doclic@tutanota.com>
+
+pkgname=<%= pkgbase %>
+pkgver=<%= version %>
+pkgrel=1
+pkgdesc="<%= description %>"
```
pdesoyres-cc, line 7: On one single repo, we now publish 2 different packages. But they will have the same description.

question: should we add the flavor description too so that user understand what is the difference between the two packages?

pdesoyres-cc (approved):
Well done @hsablonniere. I left some feedback but nothing blocking.

hsablonniere:
Thanks @pdesoyres-cc for the thorough review! Addressed all 4 points:

- docs: replaced the manual `git clone`/`makepkg` steps with `yay -S clever-tools` / `yay -S clever-tools-bin`, removed the now-redundant TIP
- CONTRIBUTING.md: trimmed the AUR section down to the essentials
- renamed the AUR variable to `AUR_BIN_GIT_URL` (GitHub Actions variable updated too)
- added the flavor label to `pkgdesc` so the two packages are distinguishable

All changes squashed into the original commits, PR marked ready for review.

hsablonniere:
Decided in #1133: the update notifier will run on every distribution by default, npm included. `--no-update-notifier` (or CI detection) disables it. The `rollup.config.js` step that strips the import for binary builds goes away.

#1133 will merge before this one, details over there.

PR #1130, opened 2026-09-02
chore(deps): drop unused @babel/traverse dependency

### LINKED ISSUE #1115: @babel/traverse is declared but never imported (16 packages in every user install)

guillaume-flambard:
## What

Removes the `@babel/traverse` dependency from the published package. It is listed under `dependencies` but never imported anywhere in the codebase. This single line was added in commit `3949a5b` alongside a codemod that has since been removed, so nothing needs it anymore.

## Why

`@babel/traverse` lands in every user install called from npm, and as a production dependency it drags in `@babel/template`, `@babel/helper-globals` and friends. Dropping it removes dead weight from every install and shrinks the dependency surface.

## Verification (on current master)

- `git grep -i traverse` matches only `package.json`; no source, script, rollup or eslint config references babel
- `npm run validate` (lint, format:check, typecheck, docs:check) passes identically before and after
- `node bin/clever.js --version` still prints `4.11.0`
- the repo `node_modules` drops 3 packages (683 to 680); the root lockfile entry was the only requester of `@babel/traverse`

Fixes #1115

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "4.11.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@babel/traverse": "7.28.5",
         "@clevercloud/client": "12.0.0",
         "@inquirer/prompts": "7.8.4",
         "char-regex": "2.0.2",
@@ -1001,6 +1000,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.28.6.tgz",
       "integrity": "sha512-JYgintcMjRiCvS8mMECzaEn+m3PfoQiyqukOMCCVQtoJGYJw8j/8LBJEiqkHLkfwCcs74E3pbAUFNg7d9VNJ+Q==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/helper-validator-identifier": "^7.28.5",
@@ -1015,6 +1015,7 @@
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.28.6.tgz",
       "integrity": "sha512-lOoVRwADj8hjf7al89tvQ2a1lf53Z+7tiXMgpZJL3maQPDxh0DgLMN62B2MKUOFcoodBHLMbDM6WAbKgNy5Suw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@babel/parser": "^7.28.6",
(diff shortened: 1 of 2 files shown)


hsablonniere:
Hello @guillaume-flambard, thanks for you PR :pray:

@HeversonSilva-gif already fixed it in #1116 so I'm going to merge it.

We're working on a 5.0.0 and it will ship with it.

PR #1133, opened 2026-09-03
fix: activate update notifier for every install channel

hsablonniere:
## Context

- The `update-notifier` message relied on an API that was removed in version 7.3.1, silently falling back to the default "npm i -g" message.
- The Rollup config strips `initial-update-notifier.js` from the bundle used for pkg binary builds, so the notifier is only available to users who install via Node.js/npm.

## Problem

- Users who install via Homebrew, WinGet, apt/yum, AUR, Exherbo or the standalone binary are never notified an update is available, even though nudging them to update is valuable.
- Users who install via pnpm, Bun or Yarn get a "npm i -g" message that doesn't match how they actually installed.

## Proposal

- Use the correct API of the latest `update-notifier` version to display a message pointing to a doc page.
- Introduce a dedicated update guide, since the install page didn't make sense once the tool is already installed.
- Promote Node.js to a top-level section in `docs/setup-systems.md`, broken into npm/pnpm/Bun/Yarn items, same treatment in the update guide.
- Remove the Rollup transform excluding the notifier from binary builds, so every install channel gets the notice.

Builds on top of #1121.

## How to test

1. Bump the local `version` in `package.json` below the latest published version (or just run against an older clone)
2. Remove the notifier cache: `rm ~/.config/configstore/update-notifier-clever-tools.json`
3. Force the check to run now: edit `~/.config/configstore/update-notifier-clever-tools.json` and set `lastUpdateCheck` to `0`, then run any `clever` command (triggers the async check)
4. Run a second `clever` command a few seconds later to display the message
5. Confirm the message and doc link, then repeat with the Rollup-built binary to confirm it also notifies

### DIFF
--- a/docs/setup-systems.md
+++ b/docs/setup-systems.md
@@ -1,7 +1,12 @@
 # How to install Clever Tools
 
-Clever Cloud CLI is based on Node.js. We thought it to be easily available on any platform. Thus, you can download Clever Tools as [a npm package](https://www.npmjs.com/package/clever-tools), but also through package managers or as a binary on many systems:
+Clever Tools is available as a npm package, through package managers, or as a standalone binary on many systems:
 
+- [Node.js](/docs/setup-systems.md#nodejs)
+  - [npm](/docs/setup-systems.md#npm)
+  - [pnpm](/docs/setup-systems.md#pnpm)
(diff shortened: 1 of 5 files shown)


hsablonniere:
@davlgd I want your feedback on these docs update.

PR #1137, opened 2026-09-09
feat: enable system git backend by default

hsablonniere:
## Context

The pure JS git implementation (isomorphic-git) has caused recurring issues: HTTP-only (no SSH), slow on repositories with rewritten history, connection timeouts on large repos, and no support for linked git worktrees.

`system-git`, the beta backend that delegates to the system-installed `git` binary, has been available as opt-in via `clever features enable system-git` and has proven reliable.

## Proposal

Flip `system-git`'s default to enabled. The feature flag stays around for a few more releases as a rollback path (`clever features disable system-git`) before eventual removal.

### Design decisions

- Each feature can now declare its own `defaultValue`, instead of `isFeatureEnabled()` hardcoding `false` for everyone
- Reworked `src/config/features.js`: user-set features are now loaded synchronously once at startup and cached in memory (`loadFeatures()`), so `isFeatureEnabled()` and the new `getAllFeatures()` no longer need to be async
- Added `getAllFeatures()`, which centralizes id/status/description/defaultValue/enabled resolution; `clever features list` now reuses it instead of duplicating that logic

## How to test

1. `clever features list` → `system-git` shows `enabled: true` without any config
2. Run any git-touching command (e.g. `clever deploy -v` from a repo linked to an app) and check the debug logs: `git(system): ...` confirms the system backend is used
3. `clever features disable system-git`, repeat step 2 → `git(isomorphic): ...` confirms the fallback still works

### DIFF
--- a/bin/clever.js
+++ b/bin/clever.js
@@ -14,7 +14,7 @@ import {
   verboseOption,
   versionOption,
 } from '../src/commands/global.options.js';
-import { EXPERIMENTAL_FEATURES, getFeatures } from '../src/config/features.js';
+import { EXPERIMENTAL_FEATURES, isFeatureEnabled } from '../src/config/features.js';
 import { cliparse } from '../src/lib/cliparse-patched.js';
 import { styleText } from '../src/lib/style-text.js';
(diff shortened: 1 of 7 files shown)


pdesoyres-cc (approved):
LGTM

PR #1143, opened 2026-09-11
fix(docs): escape pipes in generated argument tables

davlgd:
## Context

An argument whose placeholder holds a pipe breaks the markdown table it sits in. A bare pipe is a
column separator even inside backticks, so a command taking `<ng-id|ng-label>` produces this row:

```
|`ng-id|ng-label`|Network Group ID or label|
```

which GitHub renders as four columns instead of two, and the rest of the table goes out of
alignment with it. Thirteen command references were affected — every one taking something shaped
like `<ng-id|ng-label>` or `<app-id|app-name>`.

The generator already knew how to handle this: `escapeTableCell` was applied to the argument
description on that very line, and to the option aliases a few lines below. Only the argument name
was missed.

I ran into it while writing the `clever kv` documentation, but it has nothing to do with `kv`.

## Changes

Escape the argument name like everything else in those tables, and regenerate the references.

## How to review

The code change is one line, [`scripts/lib/docs-reference.js:211`](https://github.com/CleverCloud/clever-tools/blob/davlgd-docs-argument-tables/scripts/lib/docs-reference.js#L211).
The other thirteen files are `npm run docs` output.

The quickest check is visual — GitHub renders these tables, so the broken rows show up without
reading any code:

| Reference | Before | After |
|---|---|---|
| `src/commands/ng/ng.docs.md` | [master](https://github.com/CleverCloud/clever-tools/blob/master/src/commands/ng/ng.docs.md) | [this branch](https://github.com/CleverCloud/clever-tools/blob/davlgd-docs-argument-tables/src/commands/ng/ng.docs.md) |
| `src/commands/k8s/k8s.docs.md` | [master](https://github.com/CleverCloud/clever-tools/blob/master/src/commands/k8s/k8s.docs.md) | [this branch](https://github.com/CleverCloud/clever-tools/blob/davlgd-docs-argument-tables/src/commands/k8s/k8s.docs.md) |
| `src/commands/keycloak/keycloak.docs.md` | [master](https://github.com/CleverCloud/clever-tools/blob/master/src/commands/keycloak/keycloak.docs.md) | [this branch](https://github.com/CleverCloud/clever-tools/blob/davlgd-docs-argument-tables/src/commands/keycloak/keycloak.docs.md) |

### DIFF
--- a/scripts/lib/docs-reference.js
+++ b/scripts/lib/docs-reference.js
@@ -208,7 +208,7 @@ export function getSubCommandMarkdown(path, definition, subCommandNodes) {
       let description = arg.description;
       if (arg.enumValues) description += ` (${arg.enumValues.join(', ')})`;
       if (arg.optional) description += ` *${arg.optional}*`;
-      return `|\`${arg.name}\`|${escapeTableCell(description)}|`;
+      return `|\`${escapeTableCell(arg.name)}\`|${escapeTableCell(description)}|`;
     });
     parts.push(dedent`
       ### 📥 Arguments
--- a/src/commands/addon/addon.docs.md
+++ b/src/commands/addon/addon.docs.md
@@ -55,7 +55,7 @@ clever addon delete <addon-id|addon-name> [options]
 
 |Name|Description|
 |---|---|
-|`addon-id|addon-name`|Add-on ID (or name, if unambiguous)|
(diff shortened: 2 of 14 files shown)


pdesoyres-cc (approved):
LGTM

PR #1149, opened 2026-09-11
feat(ssh): select an instance by number or with any

davlgd:
Builds on #1123 by @mehdi653, whose commit is included here — the `--instance` option and its
resolution by ID are theirs.

## Context

As discussed, that option should accept instance numbers as well as IDs.

Numbers turn out not to be unique, though. During a redeploy the instance going away and the one
coming up carry the same number:

```
n°0  7e6f95ec-…  STOPPING
n°0  c320afba-…  UP
```

`--instance 0` would then reach either, and the API listed the outgoing one first.

## Changes

- Accept an instance number, and `any` in any case, alongside the ID.
- Prefer an `UP` instance: `any` looks for one before falling back to the lowest number, and a
  number prefers the `UP` instance among those sharing it. An ID still names one entry whatever its
  state, so a stopping instance stays reachable when asked for explicitly.
- List the candidates when nothing matches, instead of only saying no:

```
[ERROR] No instance 7 on this application, pick one of:
          - 0 09d1615d-629c-4278-aad2-ab475f0a4be7 (UP)
          - 1 d13eaecf-66b7-4493-8fc0-b67a933e9c25 (UP)
```

- Reject an empty value at parse time, so `--instance --command "hostname"` says what is wrong
  rather than reporting an instance named nothing.

## How to review

`selectInstance` is the whole change; the rest is the option and its documentation.

Scale an application to two instances and check that `--instance <number>` reaches the one that
number names — the remote `hostname` is the instance ID, so the output says which one answered.
Then redeploy and, while both instances share number 0, check that `--instance 0` and
`--instance any` reach the one that is `UP`.

### DIFF
--- a/src/commands/ssh/ssh.command.js
+++ b/src/commands/ssh/ssh.command.js
@@ -28,12 +28,18 @@ export const sshCommand = defineCommand({
       aliases: ['c'],
       placeholder: 'command',
     }),
+    instance: defineOption({
+      name: 'instance',
+      schema: z.string().optional(),
+      description: 'Instance ID to connect to (skips interactive selection)',
+      placeholder: 'instance-id',
+    }),
     alias: aliasOption,
     app: appIdOrNameOption,
   },
   args: [],
   async handler(options) {
-    const { alias, app: appIdOrName, identityFile, command } = options;
+    const { alias, app: appIdOrName, identityFile, command, instance } = options;
     const { appId, ownerId } = await Application.resolveId(appIdOrName, alias);
 
     const instances = await getAllInstances({ id: ownerId, appId }).then(sendToApi);
@@ -43,7 +49,13 @@ export const sshCommand = defineCommand({
     }
 
     let sshTarget;
-    if (instances.length === 1) {
+    if (instance != null) {
+      const match = instances.find((inst) => inst.id === instance);
+      if (match == null) {
+        throw new Error(`Instance ${instance} is not a running instance of this application`);
+      }
+      sshTarget = match.id;
+    } else if (instances.length === 1) {
       sshTarget = instances[0].id;
     } else if (process.stdin.isTTY) {
       const choices = instances
--- a/src/commands/ssh/ssh.command.js
+++ b/src/commands/ssh/ssh.command.js
@@ -6,6 +6,7 @@ import { config } from '../../config/config.js';
 import { defineCommand } from '../../lib/define-command.js';
 import { defineOption } from '../../lib/define-option.js';
 import { selectAnswer } from '../../lib/prompts.js';
+import { styleText } from '../../lib/style-text.js';
 import * as Application from '../../models/application.js';
 import { sendToApi } from '../../models/send-to-api.js';
 import { aliasOption, appIdOrNameOption } from '../global.options.js';
@@ -30,9 +31,9 @@ export const sshCommand = defineCommand({
     }),
     instance: defineOption({
       name: 'instance',
-      schema: z.string().optional(),
-      description: 'Instance ID to connect to (skips interactive selection)',
-      placeholder: 'instance-id',
+      schema: z.string().min(1).optional(),
+      description: 'Instance to connect to, by ID or number, or `any` (skips interactive selection)',
+      placeholder: 'instance-id|number|any',
     }),
     alias: aliasOption,
     app: appIdOrNameOption,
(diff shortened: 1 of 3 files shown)


src/commands/ssh/ssh.command.js
```diff
@@ -6,10 +6,70 @@ import { config } from '../../config/config.js';
...
+  if (wanted.toLowerCase() === 'any') {
+    return readiest(byNumber);
+  }
+
+  const found = byNumber.find((inst) => inst.id === wanted) ?? readiest(byNumber.filter(hasNumber(wanted)));
+
+  if (found == null) {
+    // Every line carries its own indent: `Logger.error` indents continuation lines by the width of
+    // its prefix, so a list indented once by the template would come out with its first line two
+    // columns further right than the others.
```
pdesoyres-cc: unecessary comments

src/commands/ssh/ssh.command.js
```diff
@@ -6,10 +6,70 @@ import { config } from '../../config/config.js';
...
+ * @returns {{ id: string, instanceNumber: number, state: string }}
+ */
+function selectInstance(instances, wanted) {
+  const byNumber = [...instances].sort((a, b) => compareNumbers(a.instanceNumber, b.instanceNumber));
+
+  if (wanted.toLowerCase() === 'any') {
+    return readiest(byNumber);
+  }
+
+  const found = byNumber.find((inst) => inst.id === wanted) ?? readiest(byNumber.filter(hasNumber(wanted)));
```
pdesoyres-cc: optimization and readability:

The `wanted` to valid number logic is made on every items of the array while filtering.
Prefer do the parsing/validation once, stop early and then filter with a valid wanted number.

src/commands/ssh/ssh.command.js
```diff
@@ -6,10 +6,70 @@ import { config } from '../../config/config.js';
...
+
+  if (found == null) {
+    // Every line carries its own indent: `Logger.error` indents continuation lines by the width of
+    // its prefix, so a list indented once by the template would come out with its first line two
+    // columns further right than the others.
+    const available = byNumber.map((inst) => `  - ${inst.instanceNumber} ${inst.id} (${inst.state})`).join('\n');
+    throw new Error(
+      `No instance ${styleText('red', wanted)} on this application, pick one of:\n${styleText('grey', available)}`,
+    );
```
pdesoyres-cc: I would have make this function just select the right instance and return null if none is found.
Throwing and error with the right message is more a command concern (should be inside the command handler function)

pdesoyres-cc (changes requested):
Few comments below. and I'm wondering if the private functions should be below the main command handler.

src/commands/ssh/ssh.command.js
davlgd: Removed

src/commands/ssh/ssh.command.js
davlgd: Now parsed once into a number, with an early return when it isn't one, so the filter only ever runs on a valid value.

src/commands/ssh/ssh.command.js
davlgd: Agreed. `selectInstance` returns `null` now and the handler raises the error, which also let the sort move up and be shared with the prompt.

davlgd:
> I'm wondering if the private functions should be below the main command handler.

Moved below the handler

pdesoyres-cc (approved):

hsablonniere:
@davlgd I rebased this on `master` and folded a few review adjustments into your commit, so it can be merged as is:

- Rebased on `master`: @mehdi653's commit is gone from the branch since it landed through #1123.
- `--instance` is now parsed by its zod schema into `any`, an instance number or an ID:
  - `toInstanceNumber()` is gone
  - `any` is matched in lowercase only, like the other enum-like values of the CLI
- `compareInstanceNumbers()` is inlined directly in the sort with a simple `?? Infinity`.
- Without `--instance`, the non-interactive error for multiple instances now lists them, like the "no instance" one. It is thrown early, before the selection, and both say "pick one with --instance". The list formatting lives in `formatInstances()`.
- Renames:
  - `ordered` → `sortedInstances` (with `toSorted()`)
  - `selected` → `selectedInstance`
  - `readiest()` → `getReadiestInstance()`
  - `inst` → `i`
- JSDoc: `instanceNumber` marked optional, to match the sort handling missing numbers.
- Commit message rewritten to state the problem.

The behaviour you described in the PR is unchanged, except `any` being case-sensitive now.

PR #1151, opened 2026-09-14
fix: add git to Dockerfile for v5

maximepvrt:
fix `[ERROR] git was not found in your PATH Either install git, or fall back to the previous JS implementation with: clever features disable system-git`in v5


hsablonniere:
Thanks for this @maximepvrt :+1:

hsablonniere:
Picked up in #1152 with a cleaner diff (consistent indentation, no touching the unrelated `ENTRYPOINT` line) 😉

PR #1152, opened 2026-09-14
fix(docker): add git dependency for system-git feature

hsablonniere:
## Summary

`system-git` (stable since v5.0.0, 3d9ebf5c) shells out to the `git` binary, which the Dockerhub image did not provide, causing:

```
[ERROR] git was not found in your PATH Either install git, or fall back to the previous JS implementation with: clever features disable system-git
```

Redo of #1151 with a clean diff (tab-consistent indentation, no touching the unrelated `ENTRYPOINT` line).

Closes #1151


maximepvrt:
@hsablonniere That doesn't work for me with `v5.0.1`. Git is only added at build time and isn't available in the final image. I think the best option might be to update the release base image (switching to `debian-slim` or `alpine`, for example) so we can install it properly.

PR #1154, opened 2026-09-14
fix(accesslogs): reconnect the stream when the API cuts it

miton18:
## Why

`clever accesslogs` ends on the first `[ERROR] Failed to receive heartbeat within 4000ms period`. That error is what a redeployment of the API looks like from the client: the instance serving the SSE stream goes away without an `END_OF_STREAM`, and the client's heartbeat watchdog (4 s, hard-coded in `@clevercloud/client`) is the only thing that notices.

`CleverCloudSse` already knows how to recover — the heartbeat timeout is a `NetworkError`, retryable, and the reconnect carries `Last-Event-ID` so the stream resumes where it stopped — but retry is opt-in (`enabled: false` by default) and `accesslogs.command.js` never opted in. `clever logs` does (`src/models/log.js`: 10 attempts, 3 s initial backoff).

## What

Pass the same `retryConfiguration` (and the same 10 s `connectionTimeout`) to `ApplicationAccessLogStream`. Retries are logged at debug level, like `clever logs`.

Context: Slack `#tech-cli`, 2026-09-14 — three `clever accesslogs` streams cut during axo.prod redeployments.


src/models/json-array.js
pdesoyres-cc, line 29: add same guard than  for open: check for `!this._isOpen` and update the is open state accordingly

pdesoyres-cc (changes requested):
Thank you @miton18 . My feedback below.

src/models/json-array.js
miton18: Done in 8678763b

pdesoyres-cc (approved):

